PR #1, opened 2018-07-19
Initial scaffold for an operator

nerd0:


Gopkg.toml
```diff
@@ -0,0 +1,39 @@
+
+# Force dep to vendor the code generators, which aren't imported just used at dev time.
```
CK-Ward: omg TG we are using `dep` 🎉

.gitignore
```diff
@@ -10,3 +10,6 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
```
prateek: avoid including `vendor/` code in the repo, makes life a lot simpler in the long run.

nerd0: Fair, I was following the practices of the private internal repositories

CK-Ward:
Are we trying to merge this anytime soon or is it a WIP?

.gitignore
prateek: Yeah fair enough. We include vendor'd deps in the internal repos for build reproducibility, so fortunately we can avoid including them upstream (GH/etc)

tmp/build/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
prateek: nit: weird to include a dir called `tmp/`, maybe separate the build scripts into a folder separate from the target output directory?

nerd0: Agreed and removed.

nerd0:
@CK-Ward This is very much a work in progress. I would like to get a single M3DB node launched with the operator before merging this in.

PR #2, opened 2018-08-02
Initial scaffold for an operator

nerd0:


Makefile
```diff
@@ -0,0 +1,18 @@
...
+build-bin: code-gen
+	@./build/build.sh
+
+.PHONY: build-docker
+build-docker: build-bin
+	@./build/docker_build.sh
+
+.PHONY: deploy
+deploy: build-docker
+	@./deploy/deployment.sh
```
CK-Ward: nit/suggestion: I usually like to have comments on each target and then a `make help` that prints each description, so it would be something like this: 

```
.PHONY: deploy
deploy: build-docker ## Builds docker image from go binary
	@./deploy/deployment.sh

.PHONY: help
help:
	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```

build/build.sh
```diff
@@ -0,0 +1,18 @@
...
+	echo "error: golang needs to be installed"
+	exit 1
+fi
+
+BIN_DIR="$(pwd)/build/output/bin"
+mkdir -p "${BIN_DIR}"
+PROJECT_NAME="m3db-operator"
+REPO_PATH="github.com/m3db/m3db-operator"
+BUILD_PATH="${REPO_PATH}/cmd/${PROJECT_NAME}"
+echo "building ${PROJECT_NAME} ..."
```
CK-Ward: I noticed there's no `go clean` in this file, or `make clean` (with `go clean` and `rm --whatever`) in the Makefile, maybe that would be useful?

build/docker_build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+if ! which docker > /dev/null; then
+	echo "docker needs to be installed"
+	exit 1
+fi
+
+: ${IMAGE:?"Need to set IMAGE, e.g. gcr.io/<repo>/<your>-operator"}
+
+GITSHA="$(git rev-parse head)"
```
CK-Ward: Do we want the `--short` flag here?

nerd0: Nope

build/build.sh
nerd0: Sure

Makefile
nerd0: Couldn't hurt, sure will add.

pkg/m3db/reconcile.go
```diff
@@ -0,0 +1,21 @@
+package m3db
+
+import (
+	api "github.com/m3db/m3db-operator/pkg/apis/m3db/v1alpha1"
+)
+
+// Reconcile reconciles the vault cluster's state to the spec specified by vr
+// by preparing the TLS secrets, deploying the etcd and vault cluster,
+// and finally updating the vault deployment if needed.
+func Reconcile(s *api.M3DBService) (err error) {
```
CK-Ward: Why does this function have a named return value for the error, but others do not?

pkg/m3db/reconcile_status.go
```diff
@@ -0,0 +1,27 @@
...
+func updateM3DBStatus(s *api.M3DBService, status *api.M3DBServiceStatus) error {
+	// don't update the status if there aren't any changes.
+	if reflect.DeepEqual(s.Status, *status) {
+		return nil
+	}
+	s.Status = *status
+	return sdk.Update(s)
+}
+
+func getM3DBStatus(s *api.M3DBService) (*api.M3DBServiceStatus, error) {
```
CK-Ward: these two functions can return errors, but I don't see error-checking implemented, is that TBD?

version/version.go
```diff
@@ -0,0 +1,5 @@
+package version
+
+var (
+	Version = "0.0.1"
```
CK-Ward: out of curiosity are we going to manually update this file or will the versions be updated via software with each green build on master?

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	@printf "    \033[36m%-15s\033[93m %s" "make" "<target>"
+	@echo
+	@echo
+	@echo "\033[92m  Targets: "
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "    \033[36m%-15s\033[93m %s\n", $$1, $$2}'
+	@printf "\033[0m"
+
+.PHONY: clean 
+clean:
+	go cleana
```
CK-Ward: I think there is a typo here? `go clean`

version/version.go
nerd0: Yes this will have to manually updated for the time being. Once we start cutting releases we can address this issue.

Makefile
nerd0: Yep thanks.

pkg/m3db/reconcile_status.go
nerd0: Currently there isn't any error-able code at the moment.

pkg/m3db/reconcile.go
nerd0: Updated, thanks for catching the inconsistency. IMHO I favor named return values but going to omit it for the time being.

nerd0:
Closing PR in favor of a [PR](https://github.com/m3db/m3db-operator/pull/3/files) that doesn't use the `operator-sdk` package.

PR #7, opened 2018-09-05
[WIP] Refactor operator controller. 

nerd0:
- Re-organizing controller, m3admin, and k8ops packages
- Added a bit more resilient logic with initial placement, namespace and stateful state operations
- Switched to gogo/protobuf and from logrus to zap
- Moved placement and namespace into separate packages. 
- Added simple status update


pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/Sirupsen/logrus"
+	"github.com/gogo/protobuf/jsonpb"
+	namespace "github.com/m3db/m3/src/dbnode/generated/proto/namespace"
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
+)
+
+const (
+	// TODO(PS) add to spec for crd?
+	// defaults for namespace creation
```
prateek: Could you use the simplified endpoints exposed for this --
```
curl -X POST http://localhost:7201/api/v1/database/create -d '{
  "type": "local",
  "namespaceName": "default",
  "retentionTime": "48h"
}'
```

 The concern is to keep these defaults in sync across multiple repos (e.g. say we add a parameter to the namespace options) is error-prone and unnecessary.

pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+	_defaultRententionOptionBufferFutureNanos                        = int64(600000000000)
+	_defaultRententionOptionBufferPastNanos                          = int64(600000000000)
+	_defaultRententionOptionBlockDataExpiry                          = true
+	_defaultRententionOptionBlockDataExpiryAfterNotAccessPeriodNanos = int64(300000000000)
+	_defaultIndexOptionenabled                                       = true
+	_defaultIndexOptionBlockSizeNanos                                = int64(7200000000000)
+)
+
+//NamespaceCreate will create a namespace
+func (m *m3admin) NamespaceCreate(cluster *myspec.M3DBCluster) error {
```
prateek: Could you create separate clients for namespace and placement?

```
type NamespaceClient interface {
  Create(...) 
  Get(...)
  Delete()
}
```

etc

pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+	_defaultRententionOptionBlockSizeNanos                           = int64(7200000000000)
+	_defaultRententionOptionBufferFutureNanos                        = int64(600000000000)
+	_defaultRententionOptionBufferPastNanos                          = int64(600000000000)
+	_defaultRententionOptionBlockDataExpiry                          = true
+	_defaultRententionOptionBlockDataExpiryAfterNotAccessPeriodNanos = int64(300000000000)
+	_defaultIndexOptionenabled                                       = true
+	_defaultIndexOptionBlockSizeNanos                                = int64(7200000000000)
+)
+
+//NamespaceCreate will create a namespace
```
prateek: nit: comments should have space between // and first char.

i.e. `// Namespace...`

pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+				BlockSizeNanos: _defaultIndexOptionBlockSizeNanos,
+			},
+		},
+	}
+	data, err := json.Marshal(namespaceCreateRequest)
+	if err != nil {
+		return err
+	}
+	// TODO(PS) - Remove hardcoded references
+	url := fmt.Sprintf("http://%s.%s:7201%s", _m3coordinatorServiceName, cluster.GetNamespace(), _namespaceURI)
```
prateek: nit: this snippet is repeated a bunch of times. Make it a helper function --

```
func (n *namespaceClient) url() string {
  ...
}
```

are you going to remove the hardcoded refs in this PR?

pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+	request, err := http.NewRequest("POST", url, bytes.NewBuffer(data))
+	if err != nil {
+		return err
+	}
+	request.Header.Add("Content-Type", "application/json")
+	if _, err := m.doHttpRequest(request); err != nil {
+		return err
+	}
+	logrus.WithField("namespace", namespaceCreateRequest.GetName()).Info("successfully created namespace")
```
prateek: We're trying to move the entire stack to use `zap` instead of other loggers. Mind using it instead of logrus?

pkg/m3admin/namespace.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, err
+	}
+	resp, err := m.doHttpRequest(request)
+	if err != nil {
+		return nil, err
+	}
+	// fix itz
+	data := &admin.NamespaceGetResponse{}
+	defer resp.Body.Close()
```
prateek: For the underlying connection to be reused, all the data from the response has to be read. Could you change this defer to the following - 

```
defer func() {
  ioutil.ReadAll(resp.Body)
  resp.Body.Close()
}()
```

pkg/m3admin/placement.go
```diff
@@ -0,0 +1,95 @@
...
+	if _, err := m.doHttpRequest(request); err != nil {
+		return err
+	}
+	logrus.Info("successfully deleted placement")
+	return nil
+}
+
+// PlacementGet will get current placement
+func (m *m3admin) PlacementGet(cluster *myspec.M3DBCluster) (*admin.PlacementGetResponse, error) {
+	url := fmt.Sprintf("http://%s.%s:7201%s", _m3coordinatorServiceName, cluster.GetNamespace(), _placementURI)
```
prateek: same feedback as other file

pkg/k8sops/statefulset.go
```diff
@@ -254,3 +257,72 @@ func (k *K8sops) BuildM3DBCoordinatorPorts() []v1.ContainerPort {
...
+					logrus.WithError(err).Error("failed to build statefulset")
+					return nil, err
+				}
+				if statefulset, err := k.Kclient.AppsV1().StatefulSets(cluster.GetNamespace()).Create(statefulset); err != nil {
+					logrus.WithError(err).WithField("statefulset", statefulset).Error("failed to create statefulset")
+				}
+				logrus.Info("statefulset created")
+
+				// Poll newly created stateful set and ensure all PODs are in ready state
+				err = wait.Poll(500*time.Millisecond, 60*time.Second, func() (bool, error) {
```
prateek: nit: use configurable knobs for these values?

pkg/k8sops/statefulset.go
```diff
@@ -254,3 +257,72 @@ func (k *K8sops) BuildM3DBCoordinatorPorts() []v1.ContainerPort {
...
 }
+
+func (k *K8sops) EnsureStatefulSets(cluster *myspec.M3DBCluster) (map[string]string, error) {
+	placementDetails := make(map[string]string) // key is pod name, value zone
+
+	// ensure zones are present in spec
+	if len(cluster.Spec.Zones) != 0 {
+		// TODO(PS) add zone distribution algo to range
+		// loop through zones
+		for zoneIndex, zoneReplicaCount := range []int32{1, 1, 1} {
```
prateek: is this going to be driven off the `cluster` object eventually?

pkg/m3admin/namespace.go
nerd0: Yep agreed, a lot of this code is in a crude state.

pkg/m3admin/namespace.go
nerd0: Good call, thanks

pkg/k8sops/statefulset.go
nerd0: Agreed, will move to the operator spec.

pkg/k8sops/statefulset.go
nerd0: yes, just updated the PR with some simple logic

nerd0:
@prateek thanks for the awesome feedback. I have addressed most of it except switching to the `database` api for the time being. 

Would you taking another pass? 

/cc @CK-Ward and @schallert

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -45,12 +45,12 @@ type M3DBClusterList struct {
...
-type CRDStatus struct {
-	State   CRDState `json:"state,omitempty"`
-	Message string   `json:"message,omitempty"`
+type M3DBStatus struct {
+	State   M3DBState `json:"state,omitempty"`
+	Message string    `json:"message,omitempty"`
 }
 
-type CRDState string
+type M3DBState string
```
prateek: might be worth making enums values if are there a defined set of values this type can take - something like:

```
type M3DBState string

const (
  GreenState M3DBState ="green"
  YellowState M3DBState ="yellow"
  RedState M3DBState ="red"
  ...
)
```

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -63,9 +63,12 @@ type ClusterSpec struct {
 	// NumberOfShards defines how many shards in total
 	NumberOfShards int32 `json:"numberOfShards"`
 
-	// Zones specifies a map of key-value pairs. Defines which zones
+	// NumberOfPods defines how many PODs should run in total
```
prateek: I'm having trouble figuring out if ClusterSpec is the desired state or the current state?

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -75,6 +78,14 @@ type ClusterSpec struct {
 	Resources Resources `json:"resources"`
 }
 
+// IsolationGroup defines the name of zone as well attributes for the zone configuration
+type IsolationGroup struct {
+	// Name
+	Name string `json:"name"`
+	// InstancesAmount defines the amount of instances
+	InstanceAmount int32 `json:"instanceAmount"`
```
prateek: What's the intent behind this value?

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
prateek: s/2016/2018/g

pkg/controller/controller.go
```diff
@@ -69,33 +74,32 @@ func (c *Controller) Init() error {
 		return err
 	}
 	if err := c.refreshClusters(); err != nil {
-		logrus.WithError(err).Error("failed to refresh clusters")
+		c.logger.Error("failed to refresh clusters", zap.Error(err))
```
prateek: cheers, thanks for making this change 👍

pkg/k8sops/common.go
```diff
@@ -0,0 +1,32 @@
...
+package k8sops
+
+import "fmt"
+
+func (k *K8sops) StatefulSetName(clusterName, isolationGroup string) string {
+	return fmt.Sprintf("%s-%s-m3", clusterName, isolationGroup)
+}
+
+const (
+	M3dbNodeSvcName      = "m3dbnode"
```
prateek: nit: casing should be `M3DB`

prateek: same for `M3CoordinatorSvcName`

pkg/k8sops/crd.go
```diff
@@ -40,26 +40,12 @@ func (k *K8sops) CreateKubernetesCustomResourceDefinition() error {
...
-					Version: m3dboperator.Version,
-					Scope:   apiextensionsv1beta1.NamespaceScoped,
-					Names: apiextensionsv1beta1.CustomResourceDefinitionNames{
-						Plural: m3dboperator.ResourcePlural,
-						Kind:   m3dboperator.ResourceKind,
-					},
-				},
-			}
-
+			k.logger.Info("CRD missing, creating it")
```
prateek: debug/trace might be a better level for this

pkg/k8sops/generators.go
```diff
@@ -0,0 +1,392 @@
...
+// GenerateService generates the Service object for m3coordinator
+func (k *K8sops) GenerateService(name string) (*v1.Service, error) {
+
+	switch name {
+	case M3dbNodeSvcName:
+		ports, err := k.GenerateServicePorts(name)
+		if err != nil {
+			return nil, err
+		}
+		return &v1.Service{
```
prateek: not familiar at all with the k8 apis, but this whole file feels like it's trying to write some config yaml in go - any way of simplifying the code?

pkg/k8sops/generators.go
```diff
@@ -0,0 +1,392 @@
...
+				Name:     "query-metrics",
+				Port:     7203,
+				Protocol: v1.ProtocolTCP,
+			},
+		}, nil
+	case M3coordinatorSvcName:
+		return []v1.ServicePort{
+			v1.ServicePort{
+				Name:     "query",
+				Port:     7201,
```
prateek: nit: should probably expose configs for all the ports to allow users to change em

pkg/k8sops/generators.go
```diff
@@ -0,0 +1,392 @@
...
+) (*appsv1.StatefulSet, error) {
+	clusterName := cluster.GetName()
+	ssName := k.StatefulSetName(clusterName, isolationGroup)
+	ports, err := k.GenerateContainerPorts(M3dbNodeSvcName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse CPU / Memory
+	limitCPU, _ := resource.ParseQuantity(clusterSpec.Resources.Limits.CPU)
```
prateek: nit: error handling for all of these Parse calls

pkg/k8sops/k8sops.go
```diff
@@ -49,15 +50,17 @@ type K8sops struct {
...
+	logger     *zap.Logger
 }
 
 // New creates a new instance of k8sops
-func New(kubeCfgFile, masterHost string) (*K8sops, error) {
+func New(logger *zap.Logger, kubeCfgFile, masterHost string) (*K8sops, error) {
 	crdClient, kubeClient, kubeExt, err := newKubeClient(kubeCfgFile)
 	if err != nil {
-		logrus.Fatalf("could not init Kubernetes client: %s", err)
+		logger.Fatal("could not init Kubernetes client", zap.Error(err))
```
prateek: It's bad practice to fatalf in functions like this - makes it harder to understand exit/crash points in the code. Could you return an error for this case instead and then if you need to fatalf the program, do so in the top level func (main/Run/whatever)

pkg/k8sops/k8sops.go
```diff
@@ -67,16 +70,15 @@ func New(kubeCfgFile, masterHost string) (*K8sops, error) {
 
 func buildConfig(kubeCfgFile string) (*rest.Config, error) {
 	if kubeCfgFile != "" {
-		logrus.Infof("using OutOfCluster k8s config with kubeConfigFile: %s", kubeCfgFile)
+		fmt.Println(fmt.Sprintf("using OutOfCluster k8s config with kubeConfigFile: %s", kubeCfgFile))
```
prateek: nit: use the logger instead?

pkg/k8sops/k8sops.go
```diff
@@ -67,16 +70,15 @@ func New(kubeCfgFile, masterHost string) (*K8sops, error) {
...
 		config, err := clientcmd.BuildConfigFromFlags("", kubeCfgFile)
 		if err != nil {
 			panic(err.Error())
```
prateek: similar to fatalf, don't panic like this either - bubble up the error to the top level

pkg/m3admin/commmon.go
```diff
@@ -0,0 +1,61 @@
...
+	"bytes"
+	"fmt"
+	"net/http"
+
+	retryhttp "github.com/hashicorp/go-retryablehttp"
+)
+
+// FormatHostname provides a formatted URL for HTTP reqeusts
+func FormatHostname(domain string) string {
+	return fmt.Sprintf("http://%s.%s:%d", M3coordinatorServiceName, domain, M3coordinatorPort)
```
prateek: probably need to make the port (service name too?) configurable externally

pkg/m3admin/commmon.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// DoHttpReqeust is a simple helper for HTTP requests
+func DoHttpRequest(
```
prateek: can you get away with this being a non-exported func?

pkg/m3admin/commmon.go
```diff
@@ -0,0 +1,61 @@
...
+	client *retryhttp.Client,
+	action, url string,
+	data *bytes.Buffer,
+) (*http.Response, error) {
+	var request *retryhttp.Request
```
prateek: unfamiliar with the retryhttp apis - is there a reason you have to break the code up into the two conditions as you have?

Would the version below work -- 
```
request, err := retryhttp.NewRequest(action, url, data)
if err != nil {
return nil, err 
}
request.Header...
...
}
```

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	"go.uber.org/zap"
+)
+
+const (
+	_httpTimeout = time.Duration(30) * time.Second
+
+	// TODO(PS) add to spec for crd?
+	// defaults for namespace creation
+	_defaultNamespaceURI                                             = "/api/v1/namespace"
```
prateek: could you expose this in m3coordinator's api package so you can reuse rather than copy paste the string

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,149 @@
...
+	_defaultName                                                     = "default"
```
prateek: for all the namespace options - could you expose a default ctor in m3db/m3coordinator and re-use across m3coord's api handlers and here

pkg/m3admin/types.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const (
+	M3coordinatorServiceName = "m3coordinator"
+	M3coordinatorPort        = 7201
+)
+
+type Namespace interface {
+	DefaultNamespaceRequest(string) *admin.NamespaceAddRequest
+	Create(string, string) (*http.Response, error)
```
prateek: some nits:
- add comments to all method/interfaces
- add argument var names -- `(string, string)` doesn't convey much about how its going to be used
- why return http.Response here? how is it going to be used externally?

pkg/k8sops/generators.go
nerd0: So I was able to convert some of the file into dynamic conversion. There a plenty with this though, if you take a look at `manifests/configmap.yaml` which contains `configmap`s of the configuration which need to be mounted at file locations in which Kubernetes is strict about unique mount points e.g. can not have the same directory mapped to a configmap which adds overhead to the operator's mounted points. Furthermore if any of these configurations contain nested configurations such as `ContainerPort` and `ServicePort` in which we would need to reconcile those two objects together. If you look at `LoadService` will show that issue since we expect configmap objects to be wholesale and not templated. IHMO it will be over optimizing the initial delivery of the operator. I agree these parameters shouldn't require a redeploy of the operator however we can make that a v2 feature.

pkg/k8sops/generators.go
nerd0: Done, but see comment above

pkg/m3admin/commmon.go
nerd0: Done, with setters.

pkg/m3admin/commmon.go
nerd0: Not sure how since this a function that is shared between two different interfaces.

pkg/m3admin/commmon.go
nerd0: The check for `nil` is due to the slice `bytes` array can be `nil` however it is still typed as a `byte` slice and the condition ensure that untyped `nil` is passed to the receiver.

pkg/m3admin/namespace/namespace.go
nerd0: Sure thing.

pkg/m3admin/namespace/namespace.go
nerd0: hmmm... isn't `m3db/m3coordinator` deprecated in favor ` the `m3db/m3` repository. Do you know the location the constants that should be used?

pkg/m3admin/types.go
nerd0: To #1 and #2 Done. 

#3 the `http.Response` is returned to allow any consumer to act on the response of the call

pkg/controller/controller.go
nerd0: cheers!

pkg/apis/m3dboperator/v1/cluster.go
nerd0: Desired state (goal state) for the cluster to converge on, see previous comment.

pkg/apis/m3dboperator/v1/cluster.go
nerd0: Yep, this was a rough commit that was recent. The idea is that there will entire status object that will define not only cluster status but node status as well. That object will be used to show the current state of the cluster holistically. If operator panics it will be able to read the status object to obtain the current state if any cluster/resource that kubernetes is aware of .

pkg/apis/m3dboperator/v1/cluster.go
nerd0: This would be the amount of instances per isolation group.

pkg/k8sops/generators.go
nerd0: Done

pkg/k8sops/k8sops.go
nerd0: Done

pkg/k8sops/k8sops.go
nerd0: Done

pkg/k8sops/k8sops.go
nerd0: Done

pkg/k8sops/generators.go
nerd0: I was able to accomplish this via the use of configmaps and marshaling into the actual structs themselves. However that approach is over complicated. After talking with @schallert , a better approach would be to have a select amount of configuration parameters exposed the spec of the M3DBCluster object and use those values if present otherwise use defaults. This would apply to service names, service ports, and container ports.

pkg/controller/controller.go
```diff
@@ -47,19 +49,30 @@ type Cluster struct {
...
-
+	placementClient, err := placement.New(logger)
+	if err != nil {
+		return nil, err
+	}
+	namespaceClient, err := namespace.New(logger)
+	if err != nil {
+		return nil, err
+	}
+	p.namespaceClient = namespaceClient
```
CK-Ward: what is the advantage of assigning this variable after the error checking, as opposed to assigning `namespace.New(logger)` to `p.namespaceClient` (and same with `p.placementClient`) right away? 

should we decide to keep the error checking & struct assigning separate, i think it would be helpful to keep a uniform order (namespace-placement-namespace-placement) instead of the arbitrary placement-namespace-namespace-placement order.

pkg/controller/controller.go
```diff
@@ -111,64 +123,46 @@ func (c *Controller) refreshClusters() error {
...
 func (c *Controller) reconcileM3DBClusterEvent(cluster *myspec.M3DBCluster) error {
 	reconcilerLock.Lock()
 	defer reconcilerLock.Unlock()
-	logrus.WithField("event", cluster.Type).Info("received M3DB cluster event")
+	if err := c.validateClusterSpec(cluster); err != nil {
+		return err
+	}
+	c.logger.Info("received M3DB cluster event", zap.String("event", cluster.Type))
 	switch {
 	case cluster.Type == "ADDED":
```
CK-Ward: indentation on `case` here please

pkg/controller/update_cluster.go
```diff
@@ -0,0 +1,74 @@
...
+	for clusterName, cluster := range c.clusters {
+		oldClusters[clusterName] = Cluster{M3DBCluster: cluster.M3DBCluster.DeepCopy()}
+	}
+
+	if err := c.refreshClusters(); err != nil {
+		c.logger.Error("failed to refresh cluster, can not compare modified state", zap.Error(err))
+		return err
+	}
+
+	currentOldClusters := oldClusters[cluster.GetName()]
```
CK-Ward: i'm a little confused on the semantic difference between `currentOldClusters` and `currentClusters`, also with the parameter just being `cluster` it doesn't really clear things up...what is the difference between "current old" vs "old"? maybe a comment to explain?

pkg/controller/util.go
```diff
@@ -0,0 +1,23 @@
+package controller
+
+import (
+	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (c *Controller) updateM3DBStatus(cluster *myspec.M3DBCluster, state string) error {
+	updatedCluster, err := c.k8sclient.CrdClient.OperatorV1().M3DBClusters(cluster.GetNamespace()).Get(cluster.GetName(), metav1.GetOptions{})
```
CK-Ward: this is a pretty long line, perhaps a clarifying comment for the reader?

pkg/k8sops/generators.go
```diff
@@ -0,0 +1,331 @@
...
+										Name: _configurationName,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
CK-Ward: lol }}}}}} yikes

pkg/m3admin/namespace.go
prateek: So the thinking behind the `database/create/` API was to provide a single endpoint with sensible defaults to make it easier for users to create a placement/namespace without thinking about the details. It basically wraps the placement/namespace APIs which have the more granular CRUD operations with some heuristics for sizing. 

How about migrating to use the create for the creation, and using the placement/namespace APIs for the more granular ops?

pkg/m3admin/types.go
prateek: Re 3 - So it's leaky to expose the `http.Response` directly. Instead of this, could you expose a native go type for this instead and the do parsing to go from http.Response -> go type within the client. 

Will make usage of the interface simpler, and be a better abstraction. For e.g. if we change to use a GRPC endpoint instead of HTTP in the future, the code using these clients won't have to change.

pkg/m3admin/namespace/namespace.go
prateek: Hm I don't think the constants exist yet. Could you use the HTTP `/database/create` endpoint (handler's at https://github.com/m3db/m3/blob/master/src/query/api/v1/handler/database/create.go)?
 probably easier than adding the constants.

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil
+	}
+}
+
+// New constructs a new namespace client
+func New(logger *zap.Logger, setters ...OptionSetter) (m3admin.Namespace, error) {
+	ns := &namespace{
+		client:      retryhttp.NewClient(),
+		logger:      logger,
+		servicePort: 7201,
```
prateek: use a const for this and the serviceName 
```
const (
  defaultServicePort = 7201
  defaultServiceName = "m3coordinator" 
)
```

pkg/k8sops/generators.go
prateek: Nice, I like it. At some point I can see us generating these configs to configure the ports or w/e, will be much easier in the structure you have now.

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -75,6 +78,14 @@ type ClusterSpec struct {
...
+}
+
 // Resources defines CPU / Memory restrictions on pods
```
prateek: Could you add a comment explaining difference between `Requests` v `Limits`

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -63,9 +63,12 @@ type ClusterSpec struct {
 	// NumberOfShards defines how many shards in total
 	NumberOfShards int32 `json:"numberOfShards"`
 
-	// Zones specifies a map of key-value pairs. Defines which zones
+	// NumberOfPods defines how many PODs should run in total
+	NumberOfPods int32 `json:"numberOfPods"`
+
+	// Zones specifies a map of key-value pairs. Defines which isolation groups
```
prateek: comment needs to be updated to reflect new name

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -75,6 +78,14 @@ type ClusterSpec struct {
...
+// IsolationGroup defines the name of zone as well attributes for the zone configuration
+type IsolationGroup struct {
+	// Name
+	Name string `json:"name"`
+	// InstancesAmount defines the amount of instances
+	InstanceAmount int32 `json:"instanceAmount"`
+}
+
 // Resources defines CPU / Memory restrictions on pods
 type Resources struct {
```
prateek: Should `Disk` be included in this?

prateek: could you constrain the types for Memory and CPU limits (to something more specific than `string`)

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
...
+const (
+	M3coordinatorServiceName = "m3coordinator"
+	M3coordinatorPort        = 7201
+)
+
+type Namespace interface {
+	// DefaultNamespaceRequest Creates a Namespace Add Request with a given namespace
+	DefaultNamespaceRequest(namspace string) *admin.NamespaceAddRequest
+	// Create will create a namepace with provided namespace name and defaults
+	Create(hostname string, namespace string) (*http.Response, error)
```
prateek: what's the hostname for?

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	"github.com/m3db/m3cluster/generated/proto/placementpb"
+)
+
+const (
+	M3coordinatorServiceName = "m3coordinator"
+	M3coordinatorPort        = 7201
+)
+
+type Namespace interface {
```
prateek: nit: probably better to call this `NamespaceClient`

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
...
+	Create(hostname string, namespace string) (*http.Response, error)
+	// Get will retrieve a namespace given a name
+	Get(hostname string) (*admin.NamespaceGetResponse, error)
+	// Delete will delete a namespace given a name
+	Delete(hostname string, namespace string) (*http.Response, error)
+	// FormatHostname will provide a valid hostname string to be used with requests
+	FormatHostname(domain string) string
+}
+
+type Placement interface {
```
prateek: same as comments for `NamespaceClient` ...

pkg/m3admin/types.go
prateek: Could do this with a struct slightly better - 

```
package namespace 

type nsClient struct {
  hostname string 
}

func NewClient(hostname string) (NamespaceClient, error) {
  // set hostname in the struct
  // and you won't have to keep passing around hostname in the different methods
}
```

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	M3coordinatorServiceName = "m3coordinator"
+	M3coordinatorPort        = 7201
+)
+
+type Namespace interface {
+	// DefaultNamespaceRequest Creates a Namespace Add Request with a given namespace
+	DefaultNamespaceRequest(namspace string) *admin.NamespaceAddRequest
```
prateek: why expose `DefaultNamespaceRequest`?

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
...
+	// Create will create a namepace with provided namespace name and defaults
+	Create(hostname string, namespace string) (*http.Response, error)
+	// Get will retrieve a namespace given a name
+	Get(hostname string) (*admin.NamespaceGetResponse, error)
+	// Delete will delete a namespace given a name
+	Delete(hostname string, namespace string) (*http.Response, error)
+	// FormatHostname will provide a valid hostname string to be used with requests
+	FormatHostname(domain string) string
```
prateek: don't have to expose this either.

pkg/m3admin/types.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
prateek: s/2016/2018/g

pkg/apis/m3dboperator/v1/cluster.go
prateek: ha ok, how about renaming to `NumInstances`?

pkg/m3admin/placement/placement.go
```diff
@@ -0,0 +1,171 @@
...
+	"go.uber.org/zap"
+)
+
+var ErrPlacementNotFound = errors.New("placement not found")
+
+const (
+	_defaultPlacementURI = "/api/v1/placement"
+
+	// defaults for placement init request
+	DefaultZoneType = "embedded"
```
prateek: do these have to be exported?

nerd0:
Addressed most of the comments, the next PR will address the comments missed: 

- Import m3 packages beside generated (protobuf) or util packages. Currently there are dependencies issues that I will need to sort out before doing so. 

- Change creation of namespace and placement to the database api 

- Move namespace and placement clients within Cluster object within the controller to allow multiple clusters to maintain different clients 

- Handle instance increase and decrease changes to the spec file. 

- Port validation introspecting the m3.yaml configuration to ensure align with m3 configuration and kubernetes configuration with respect to port allocation. 

- Switch from stateful sets to pod based management

- Handle deployments of new m3 docker images.

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -45,14 +45,20 @@ type M3DBClusterList struct {
 	Items           []M3DBCluster `json:"items"`
 }
 
-type CRDStatus struct {
-	State   CRDState `json:"state,omitempty"`
-	Message string   `json:"message,omitempty"`
+type M3DBStatus struct {
```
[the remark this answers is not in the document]
nerd0: Good catch, currently linting isn't enabled for the project. I am in the middle of setting up all the CI bits as well.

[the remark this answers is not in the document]
nerd0: Ran linter and fixed all linting issues that popped up, thanks again.

Makefile
```diff
@@ -6,36 +6,125 @@ BUILD_SETTINGS := GOOS=linux GOARCH=amd64 CGO_ENABLED=0
...
+clean:
+	@rm -f *.html *.xml *.out *.test
+	@go clean
+	@rm -f build/output/bin/m3db-operator
+
+.PHONY: all
+all: clean code-gen lint metalint test-ci-unit
+	@echo make all successfully finished
+
+#.PHONY: help
```
prateek: nit: delete this

prateek:
@nerd0 wanna setup semaphore/travis too?

pkg/m3admin/namespace/types.go
```diff
@@ -0,0 +1,33 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package namespace
+
+import "github.com/m3db/m3/src/query/generated/proto/admin"
+
+// Namespace provides the interface to interact with the namespace API
+type Namespace interface {
```
prateek: nit: maybe just call this `Client` - that way it'll be `namespace.NewClient` or w/e the ctor is

pkg/m3admin/placement/placement.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/gogo/protobuf/jsonpb"
+	retryhttp "github.com/hashicorp/go-retryablehttp"
+	plc "github.com/m3db/m3/src/query/api/v1/handler/placement"
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	"github.com/m3db/m3/src/query/util/logging"
+	"go.uber.org/zap"
+)
+
+// ErrPlacementNotFound indicates that the placement is not found
+var ErrPlacementNotFound = errors.New("placement not found")
```
prateek: nit: why make this exported?

pkg/m3admin/placement/placement.go
```diff
@@ -0,0 +1,197 @@
...
+func WithLogger(logger *zap.Logger) Option {
+	return optionFn(func(p *placement) error {
+		p.logger = logger
+		return nil
+	})
+}
+
+// New is the constructor the Placement interface
+func New(opts ...Option) (Placement, error) {
+	logging.InitWithCores(nil)
```
prateek: why initialise logging here? 

would be better to use a https://godoc.org/go.uber.org/zap#NewNop here by default and let users override with the option if so desired

pkg/m3admin/placement/types.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	"github.com/m3db/m3cluster/generated/proto/placementpb"
+)
+
+// Placement provides the interface to interact with the placement API
+type Placement interface {
+	// Init will initialize a placement give a valid placement request
+	Init(request *admin.PlacementInitRequest) error
```
prateek: hm could you use golang native types from https://github.com/m3db/m3cluster/blob/master/placement/types.go#L109 instead of the proto types?

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,218 @@
...
+func WithLogger(logger *zap.Logger) Option {
+	return optionFn(func(n *namespace) error {
+		n.logger = logger
+		return nil
+	})
+}
+
+// New constructs a new namespace client
+func New(opts ...Option) (Namespace, error) {
+	logging.InitWithCores(nil)
```
prateek: same as logger feedback in `placement`

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	_, err = m3admin.DoHTTPRequest(n.client, "POST", url, bytes.NewBuffer(data))
+	if err != nil {
+		return err
+	}
+	n.logger.Info("successfully created namespace", zap.String("namespace", namespaceName))
+	return nil
+}
+
+// Get will retrieve a namespace
```
prateek: which namespace will it retrieve? if it's all of them - then this it's probably better to call this `List()` 

also, could you use golang native types we've exposed - e.g. `[]string` for namespace names, or a slice of https://github.com/m3db/m3/blob/master/src/dbnode/storage/namespace/types.go#L108

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,218 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"time"
+
+	"github.com/m3db/m3db-operator/pkg/m3admin"
+
+	"github.com/gogo/protobuf/jsonpb"
```
prateek: so import convention we follow for OSS projects:

```
import (
   "stdlib"
   
   "all-imports-starting with 'github.com/m3db/'" 
 
   "other imports"
)
```

would be good to wire up https://github.com/m3db/m3/blob/master/src/cmd/.metalinter.json#L15 to catch this kinda thing

pkg/m3admin/commmon.go
prateek: hm could you avoid this issue if you just expected `data []byte` instead of `data *bytes.Buffer`?

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+		Message: "m3 db is initializing",
+	}
+	if err := c.updateM3DBStatus(cls.M3DBCluster, status); err != nil {
+		return err
+	}
+	// create services defined in service configuration
+	for _, svcCfg := range cls.M3DBCluster.Spec.ServiceConfigurations {
+		if err := c.k8sclient.EnsureService(cls.M3DBCluster, svcCfg); err != nil {
+			redStatus.Message = fmt.Sprintf("%s: %v", redStatus.Message, err)
+			if err := c.updateM3DBStatus(cls.M3DBCluster, redStatus); err != nil {
```
prateek: can simplify this to `return c.updateM3DBStatus(..)`

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+
+	// TODO(PS) remove loop to find the correct service
+	for _, svcCfg := range cls.M3DBCluster.Spec.ServiceConfigurations {
+		if svcCfg.Name == "m3dbnode" {
+			// TODO(PS) replace statefulsets with pods instead
+			if err := c.k8sclient.EnsureStatefulSets(cls.M3DBCluster, svcCfg); err != nil {
+				redStatus.Message = fmt.Sprintf("%s: %v", redStatus.Message, err)
+				if err := c.updateM3DBStatus(cls.M3DBCluster, redStatus); err != nil {
```
prateek: here too

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+					return err
+				}
+				return err
+
+			}
+		}
+	}
+	if err := c.EnsureNamespace(cls.M3DBCluster); err != nil {
+		redStatus.Message = fmt.Sprintf("%s: %v", redStatus.Message, err)
+		if err := c.updateM3DBStatus(cls.M3DBCluster, redStatus); err != nil {
```
prateek: here too

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+			return err
+		}
+		return err
+	}
+	if err := c.EnsurePlacement(cls.M3DBCluster); err != nil {
+		redStatus.Message = fmt.Sprintf("%s: %v", redStatus.Message, err)
+		if err := c.updateM3DBStatus(cls.M3DBCluster, redStatus); err != nil {
```
prateek: here too

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+	//
+	// TODO(PS) Ensure all namepaces can be validated in a registry
+	namespaces := resp.GetRegistry().GetNamespaces()
+	if _, ok := namespaces[cluster.GetObjectMeta().GetName()]; ok {
+		c.logger.Info("namespace found")
+		return nil
+	}
+
+	if err = c.namespaceClient.Create(cluster.GetObjectMeta().GetName()); err != nil {
+		c.logger.Error("failed to create namespace after 6 tries", zap.Error(err))
```
prateek: where do `6 tries` come from?

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+	resp, err := c.namespaceClient.Get()
+	if err != nil {
+		c.logger.Error("failed to get namespace ", zap.Error(err))
+		return err
+	}
+
+	// check if namespace already exist
+	//
+	// TODO(PS) Ensure all namepaces can be validated in a registry
+	namespaces := resp.GetRegistry().GetNamespaces()
```
prateek: this is what i mean about namespace client returning native types - would allow the logic here to be a bit simpler - you'd be guaranteed the Registry/Metadatas are valid by this point.

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,167 @@
...
+	if _, ok := namespaces[cluster.GetObjectMeta().GetName()]; ok {
+		c.logger.Info("namespace found")
+		return nil
+	}
+
+	if err = c.namespaceClient.Create(cluster.GetObjectMeta().GetName()); err != nil {
```
prateek: hm so a m3db cluster can have more than one namespace. does the spec need to be updated to detect that

Makefile
nerd0: done

pkg/m3admin/namespace/types.go
nerd0: Sure thing, done

pkg/m3admin/placement/placement.go
nerd0: To help the consumer check for this type of error.

pkg/m3admin/placement/placement.go
nerd0: Great call! Done.

pkg/m3admin/namespace/namespace.go
nerd0: Ditto, Done.

pkg/apis/m3dboperator/register.go
```diff
@@ -27,14 +27,23 @@ import (
...
-	ResourceKind   = "M3DBCluster"
-	ResourcePlural = "m3dbclusters"
-	GroupName      = "operator.m3db.io"
-	ShortName      = "m3dbcluster"
-	Version        = "v1"
+	// ResourceKind is the custom resource kind
+	ResourceKind = "M3DBCluster"
+
+	// ResourcePlural is the plural form of custom resource kind
+	ResourcePlural = "M3DBclusters"
```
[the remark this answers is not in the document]
nerd0: Good catch, done.

pkg/apis/m3dboperator/register.go
```diff
@@ -27,14 +27,23 @@ import (
...
+
+	// GroupName is the group that the custom resource belongs to
+	GroupName = "operator.m3db.io"
```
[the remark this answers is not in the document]
nerd0: It's really all over the place ATM, this ensures that the fully qualified name doesn't collide with other operators. The convention I have seen is <controller_purpose>.<company_website>

pkg/apis/m3dboperator/v1/zz_generated.deepcopy.go
```diff
@@ -9,39 +9,39 @@ import (
 )
 
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
-func (in *CRDStatus) DeepCopyInto(out *CRDStatus) {
+func (in *ClusterSpec) DeepCopyInto(out *ClusterSpec) {
```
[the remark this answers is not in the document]
nerd0: There is a target in the `Makefile` for `code-gen`

pkg/controller/add_cluster.go
nerd0: Hmmm... then the `error` from `EnsureService` will be lost?

pkg/controller/add_cluster.go
nerd0: ditto...then the `error` from `EnsureService` will be lost?

pkg/controller/add_cluster.go
nerd0: ditto...then the `error` from `EnsureService` will be lost?

pkg/controller/add_cluster.go
nerd0: old string that didn't get updated, fixed it.

pkg/controller/add_cluster.go
nerd0: Fair point and I agree.

pkg/m3admin/namespace/namespace.go
nerd0: Yep, and going to wire metalinter soon as well.

pkg/m3admin/namespace/namespace.go
nerd0: Updated it.

pkg/m3admin/commmon.go
nerd0:
```
// NewRequest creates a new wrapped request.                                   
func NewRequest(method, url string, rawBody interface{}) (*Request, error) {   
  var err error                                                                
  var body ReaderFunc                                                          
  var contentLength int64                                                         
                                                                                  
  if rawBody != nil {                                                             
    switch rawBody.(type) {  
....
```

Not sure, it's an untyped `interface` that is evaluated within a `switch` for several variants in which `[]byte` is one of the types that is cased.

nerd0:
@prateek got semaphore, codecov, and coveralls setup. However since I set them after this branch was created those CI tools aren't picking up these changes. I hazard that after this lands those tools will kick in. The `make` targets are in place to take care of the reports as well but was only able to validate `coveralls`, not sure why `codecov` didn't work.

nerd0:
@prateek put alot of happy path test cases. Feel free to comment

pkg/m3admin/placement/placement.go
prateek: Fair enough, presumably they need to act upon it differently from other errors? if so, maybe add blurb comment to the method that returns this as well

pkg/m3admin/placement/placement.go
```diff
@@ -0,0 +1,189 @@
...
+	return optionFn(func(p *placement) error {
+		p.logger = logger
+		return nil
+	})
+}
+
+// NewClient is the constructor the Placement interface
+func NewClient(opts ...Option) (Client, error) {
+	logger := zap.NewNop()
+	// TODO(PS) Add logger.Sync() somewhere
```
prateek: nit: remove this now

pkg/m3admin/placement/types.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/m3db/m3/src/query/generated/proto/admin"
+	"github.com/m3db/m3cluster/generated/proto/placementpb"
+)
+
+// Client provides the interface to interact with the placement API
+type Client interface {
+	// Init will initialize a placement give a valid placement request
+	Init(request *admin.PlacementInitRequest) error
+	// Get will provide a placement given a hostname e.g. a subdomin
```
prateek: comments need to be updated to indicate a hostname isn't required.

pkg/m3admin/placement/types.go
```diff
@@ -0,0 +1,38 @@
...
+	Get() (*admin.PlacementGetResponse, error)
+	// Delete will delete a placment given a hostname e.g. a subdomain
+	Delete() error
+	// Add will add a placement given a hostname e.g. a subdomain
+	Add(instance placementpb.Instance) (*admin.PlacementGetResponse, error)
```
prateek: could you use native types instead of proto types

pkg/m3admin/namespace/types.go
```diff
@@ -0,0 +1,33 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package namespace
+
+import ns "github.com/m3db/m3/src/dbnode/storage/namespace"
+
+// Client provides the interface to interact with the namespace API
+type Client interface {
```
prateek: nice

pkg/m3admin/namespace/namespace.go
```diff
@@ -0,0 +1,223 @@
...
+	return optionFn(func(n *namespace) error {
+		n.logger = logger
+		return nil
+	})
+}
+
+// NewClient constructs a new namespace client
+func NewClient(opts ...Option) (Client, error) {
+	logger := zap.NewNop()
+	// TODO(PS) Add logger.Sync() somewhere
```
prateek: nit: delete this line

Makefile
```diff
@@ -6,36 +6,125 @@ BUILD_SETTINGS := GOOS=linux GOARCH=amd64 CGO_ENABLED=0
...
+	gocov convert $(coverfile) | gocov-html > $(html_report) && open $(html_report)
+	@rm -f $(test_log) &> /dev/null
+
+.PHONY: test-ci-unit
+test-ci-unit: test-internal
+	@which goveralls > /dev/null || go get -u -f github.com/mattn/goveralls
+	goveralls -coverprofile=$(coverfile) -service=travis-ci || echo -e "\x1b[31mCoveralls failed\x1b[m"
+
+.PHONY: install-mockgen
+install-mockgen: install-vendor-dep
```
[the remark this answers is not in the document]
prateek: Yeah i've seen that and I wonder about it. 

There are definitely tradeoffs - e.g. it'd save you a bunch of hassle longer term 
- versioning/releasing/docs/etc would all be bundled together; 
- use APIs (m3coordinator/m3db/etc) without worrying about versions being pinned correctly 
- imagine if you have to create operators for other components - probably don't want to make separate repos for each one.

definite downsides:
- the monorepo build time and test flakiness aren't great at the moment. 

Brought this up as more something to think about. See how things go while dev'ing, can circle back before release

cmd/m3db-operator/main.go
```diff
@@ -21,19 +21,27 @@
...
 	"flag"
-	"fmt"
 	"os"
 	"os/signal"
 	"runtime"
 	"sync"
 	"syscall"
 
+	"github.com/m3db/m3/src/query/util/logging"
+	clientset "github.com/m3db/m3db-operator/pkg/client/clientset/versioned"
```
prateek: why two aliases for the same package?

cmd/m3db-operator/main.go
```diff
@@ -46,45 +54,49 @@ var (
...
+
+	logger.Info("Go", zap.Any("VERSION", runtime.Version()))
+	logger.Info("Go", zap.Any("OS", runtime.GOOS), zap.Any("ARCH", runtime.GOARCH))
+	logger.Info("Operator", zap.String("version", appVersion))
+
+	// Create k8s clients
+	crdClient, kubeClient, kubeExt, err := newKubeClient(logger, kubeCfgFile)
 	if err != nil {
-		fmt.Println("failed to setup log level correctly")
+		logger.Error("failed to create k8s clients", zap.Error(err))
```
prateek: nit: can simplify to `logger.Fatal(...)`

cmd/m3db-operator/main.go
```diff
@@ -46,45 +54,49 @@ var (
...
+	// Create k8sops client
+	k8sclient, err := k8sops.New(
+		masterHost,
+		k8sops.WithLogger(logger),
+		k8sops.WithCRDClient(crdClient),
+		k8sops.WithKClient(kubeClient),
+		k8sops.WithExtClient(kubeExt))
 	if err != nil {
-		logrus.WithError(err).Error("failed to create k8sclient")
+		logger.Error("failed to create k8sclient", zap.Error(err))
```
prateek: in all these spots

pkg/controller/add_cluster.go
prateek: yeah, fair enough

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+const (
+	// defaults for placement init request
+	_defaultM3DBPort = 9000
+)
+
+func (c *Controller) addM3DBCluster(cluster *myspec.M3DBCluster) error {
+
+	redStatus := myspec.M3DBStatus{
+		State:   myspec.RedState,
+		Message: "m3 db failed to initialize",
```
prateek: use consts for this kinda messaging

pkg/controller/add_cluster.go
prateek: yep fair.

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+			if err := c.updateM3DBStatus(cls.M3DBCluster, redStatus); err != nil {
+				return err
+			}
+			return err
+		}
+	}
+
+	// TODO(PS) remove loop to find the correct service
+	for _, svcCfg := range cls.M3DBCluster.Spec.ServiceConfigurations {
+		if svcCfg.Name == "m3dbnode" {
```
prateek: hm maybe use a function to get a svc's config from the spec, it'd simplify the code a bunch, e.g. 

```
func getServiceConfig(svcConfigs ..., serviceName string) (config, bool) {
  for _, svcConfig := range svcConfigs { 
    if svcConfig.Name == serviceName {
      return svcConfig, true
    }
  }
  return config{}, false
}
```

pkg/controller/add_cluster.go
```diff
@@ -0,0 +1,165 @@
...
+	status = myspec.M3DBStatus{
+		State:   myspec.GreenState,
+		Message: "m3 db is up and running",
+	}
+	return c.updateM3DBStatus(cls.M3DBCluster, status)
+}
+
+// EnsurePlacement ensures that a placement exists otherwise create one
+func (c *Controller) EnsurePlacement(cluster *myspec.M3DBCluster) error {
+	//get placement
```
prateek: nit: space after comment

pkg/controller/controller.go
```diff
@@ -108,186 +121,66 @@ func (c *Controller) refreshClusters() error {
 	//Reset
 	c.clusters = make(map[string]Cluster)
 
-	// Get existing clusters
-	currentClusters, err := c.k8sclient.CrdClient.OperatorV1().M3DBClusters(v1.NamespaceAll).List(metav1.ListOptions{})
+	//Get active clusters
```
prateek: nit space in the comment, should be
``` // Get``` not ```//Get```

File: pkg/controller/controller.go [hunk 77]
```diff
@@ -108,186 +121,66 @@ func (c *Controller) refreshClusters() error {
...
-		if err := c.deleteM3DBCluster(cluster); err != nil {
-			return err
-		}
-		if err := c.k8sclient.DeleteM3DBNodeSvc(cluster); err != nil {
+func (c *Controller) validateClusterSpec(cluster *myspec.M3DBCluster) error {
+	// TODO(PS) LoadM3Configuration when dep conflicts are resolved. The
+	// configuration will allow us to validate the port numbers specified in the
+	// m3 configuration align with the ones within the M3DBCluster spec.
+	/*
+		m3Config, err := c.LoadM3Configuration(cluster)
```
prateek: nit: delete commented out code before landing

pkg/controller/controller.go
```diff
@@ -108,186 +121,66 @@ func (c *Controller) refreshClusters() error {
...
-		statefulsetName := c.k8sclient.StatefulsetName(cluster.ObjectMeta.Name, zone, 1)
-		logrus.WithField("statefulsetName", statefulsetName).Info("deleting stateful set")
-		if err := c.k8sclient.Kclient.AppsV1().
-			StatefulSets(cluster.GetNamespace()).
-			Delete(statefulsetName, &metav1.DeleteOptions{}); err != nil {
-			logrus.WithError(err).WithField("statefulsetName", statefulsetName).Error("failed to delete statefulset")
-		}
-	}
+func (c *Controller) deployNewImage(cluster *myspec.M3DBCluster) error {
+	//_, _ = c.k8sclient.Kclient.AppsV1().Statefulsets(cluster.GetNamespace()).Update()
```
prateek: here too

pkg/controller/delete_cluster.go
```diff
@@ -0,0 +1,45 @@
...
+	if err := c.namespaceClient.Delete(cluster.GetObjectMeta().GetName()); err != nil {
+		return err
+	}
+	if err := c.placementClient.Delete(); err != nil {
+		return err
+	}
+	if err := c.k8sclient.DeleteStatefulSets(cluster, c.k8sclient.LabelSelector("cluster", cluster.GetName())); err != nil {
+		return err
+	}
+	if err := c.k8sclient.DeleteService(cluster, "m3dbnode"); err != nil {
```
prateek: nit: use consts for service names that can be shared across packages (or use config).

pkg/k8sops/generators.go
```diff
@@ -0,0 +1,307 @@
...
+		FailureThreshold:    _probeFailureThreshold,
+		Handler: v1.Handler{
+			HTTPGet: &v1.HTTPGetAction{
+				Port:   intstr.FromInt(_probePort),
+				Path:   _probePath,
+				Scheme: v1.URISchemeHTTP,
+			},
+		},
+	}
+	statefulset := &appsv1.StatefulSet{
```
prateek: can you use the same hackery as the other spot to make this a little cleaner?

pkg/k8sops/types.go
```diff
@@ -0,0 +1,128 @@
...
+	CreateStatefulSet(cluster *myspec.M3DBCluster, statefulSet *appsv1.StatefulSet) (*appsv1.StatefulSet, error)
+
+	// GetStatefulSet simply returns a StatefulSet given the current cluster
+	GetStatefulSet(cluster *myspec.M3DBCluster, name string) (*appsv1.StatefulSet, error)
+
+	// UpdateStatefulSet simply updates a statefulset
+	UpdateStatefulSet(cluster *myspec.M3DBCluster, statefulSet *appsv1.StatefulSet) (*appsv1.StatefulSet, error)
+
+	// CheckStatefulStatus will poll a given StatefulSet to ensure it reaches a
+	// ready state within 60 seconds with polling updates at 500 ms
```
prateek: probably best to avoid specific numbers here and say "within a configurable period" or something

pkg/k8sops/types.go
```diff
@@ -0,0 +1,128 @@
...
+	// GenerateServicePorts will produce the correct ServicePort or ContainerPort
+	// resources
+	GenerateServicePorts(ports []myspec.Port) []v1.ServicePort
+
+	// GenerateContainerPorts will produce ServicePorts given a
+	// ServiceConfiguration
+	GenerateContainerPorts(ports []myspec.Port) []v1.ContainerPort
+
+	// MonitorM3DBEvents watches for new or removed clusters
+	MonitorM3DBEvents(stopchan chan struct{}) (<-chan *myspec.M3DBCluster, <-chan error)
```
prateek: Go channels work in weird ways (e.g. sending on a blocked channel can cause a panic), it's frowned upon to use channels across interface boundaries. Could you use something like https://github.com/m3db/m3x/blob/master/watch/watch.go to wrap the types being passed around.

It'll give you much clearer expectations about ownership - who has to close which object and so on.

pkg/k8sops/types.go
```diff
@@ -0,0 +1,128 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package k8sops
+
+import (
+	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
```
prateek: nit: import order

prateek:
@nerd0 maybe break out the bits relevant to CI in another PR and land that first. Then rebase this to ensure it's all good? 

re: `codecov` v `coveralls` - feel free to use whichever one you prefer. We migrated to `codecov` from `coveralls` a while back in other repos because it gave us more control over submitting multiple reports for a single CI job and performed better on the UI.

prateek:
code's looking much better, had far fewer comments this pass 👍

pkg/k8sops/types.go
nerd0: Yea ended up moving this within the Controller package and handling the channels within the package itself

nerd0:
[PR](https://github.com/m3db/m3db-operator/pull/8) for CI bits. Currently failing lint but so much of that code has changed, not it's worth revisiting those issues?

pkg/k8sops/generators.go
nerd0: Actually I want to move to another resource type in the next diff. Could we hold off on those changes till then?

nerd0:
Closing in favor of [PR #9](https://github.com/m3db/m3db-operator/pull/9)

PR #8, opened 2018-09-17
Adding ci bits

nerd0:


pkg/m3admin/m3admin_test.go
```diff
@@ -0,0 +1 @@
+package m3admin
```
prateek: nit: license header in every go file

prateek:
@nerd0 looks like build is :( too

pkg/m3admin/m3admin_test.go
nerd0: done

PR #11, opened 2018-09-26
model event loop w/ workqueue; pin sets to zones

schallert:
- Model the event loop using a client-go's workqueue. This lets us have many workers safely processing events and retry events with backoff. Using this abstraction may also make testing easier.

- Pin each statefulset to a zone using the pod spec's affinity. A new statefulset isn't brought up until the previous one is ready, and nothing happens until all statefulsets are ready (we'll have to inject placement recovery before this later on).

- Begin using `[]Conditions` pattern in status.

- Add a command-line override to allow communicating w/ coordinator over localhost tunnel.

Have used this to get a cluster up and running spread across 3 zones with namespace initialized.


pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -52,9 +66,60 @@ type M3DBStatus struct {
...
+	Conditions []ClusterCondition `json:"conditions,omitempty"`
+
 	// Message is a human readable message indicating why the cluster is in it's
 	// current state
 	Message string `json:"message,omitempty"`
+
+	// ObservedGeneration is the last generation of the cluster the controller
+	// observed. Kubernetes will automatically increment metadata.Generation every
+	// time the cluster spec is changed.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
nerd0: Nice

pkg/controller/add_cluster.go
```diff
@@ -170,24 +172,61 @@ func (c *Controller) EnsurePlacement(cluster *myspec.M3DBCluster) error {
...
 	}
 	if err = c.namespaceClient.Create(cluster.GetObjectMeta().GetName()); err != nil {
 		c.logger.Error("failed to create namespace", zap.Error(err))
+		return false, err
+	}
+	return true, nil
+}
+
+func (c *Controller) ensureServices(cluster *myspec.M3DBCluster) error {
+	// TODO(schallert): support updating service spec, not sure if this only
```
nerd0: Yea that is fair because the heuristic is a simple check that the service exist or not against the entire service configuration such as ports.

pkg/controller/controller.go
```diff
@@ -22,17 +22,47 @@ package controller
...
+
 	"github.com/m3db/m3db-operator/pkg/k8sops"
 	"github.com/m3db/m3db-operator/pkg/m3admin/namespace"
 	"github.com/m3db/m3db-operator/pkg/m3admin/placement"
 
 	"go.uber.org/zap"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
```
nerd0: nit: remove line

pkg/controller/controller.go
```diff
@@ -22,17 +22,47 @@ package controller
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+
```
nerd0: nit: remove line

pkg/controller/controller.go
```diff
@@ -22,17 +22,47 @@ package controller
...
 	"sync"
 	"time"
 
 	"github.com/m3db/m3db-operator/pkg/apis/m3dboperator"
 	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
+	clientset "github.com/m3db/m3db-operator/pkg/client/clientset/versioned"
+	samplescheme "github.com/m3db/m3db-operator/pkg/client/clientset/versioned/scheme"
+	informers "github.com/m3db/m3db-operator/pkg/client/informers/externalversions"
+	clusterlisters "github.com/m3db/m3db-operator/pkg/client/listers/m3dboperator/v1"
+
```
nerd0: nit: remove line

pkg/controller/controller.go
```diff
@@ -125,9 +241,286 @@ func (c *Controller) Start(wg *sync.WaitGroup) error {
...
+			Version: myspec.SchemeGroupVersion.Version,
+			// TODO(schallert): use a const here
+			Kind: "m3dbcluster",
+		})
+
+		name := fmt.Sprintf("%s-%d", cluster.Name, nextID)
+		sts := newBaseStatefulSet(name, cluster.Spec.Image, cluster.Name, zones[nextID], 1)
+
+		// Create an Affinity policy for isolationgroup.
+		nodeAffinity := &corev1.NodeAffinity{
```
nerd0: Make this a function if possible.

pkg/controller/controller.go
```diff
@@ -125,9 +241,286 @@ func (c *Controller) Start(wg *sync.WaitGroup) error {
...
+		}
+
+		c.logger.Info("recovered object from tombstone", zap.String("name", object.GetName()))
+	}
+
+	c.logger.Info("processing statefulset", zap.String("name", object.GetName()))
+
+	owner := metav1.GetControllerOf(object)
+	// TODO(schallert): const
+	if owner == nil || owner.Kind != "m3dbcluster" {
```
nerd0: isn't `Kind` named `M3DBCluster`, however to be honest I wasn't very much a fan of that but code `M3` using `M3DB` everywhere. Naming is hard in programming.

cmd/m3db-operator/main.go
```diff
@@ -83,35 +94,59 @@ func main() {
...
+	// them
+	//
+	// TODO(schallert): not sure if we need podinformers as well or if
+	// abstractions of statefulsets will do
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, _informerSyncDuration)
+	m3dbClusterInformerFactory := informers.NewSharedInformerFactory(crdClient, _informerSyncDuration)
+
 	// Create controller
-	controller, err := controller.New(logger, k8sclient)
+	controller, err := controller.New(logger, k8sclient, crdClient, kubeClient, kubeInformerFactory, m3dbClusterInformerFactory)
```
nerd0: should the controller have directly access to the `crdClient` and `kubeClient`?

pkg/controller/add_cluster.go
```diff
@@ -170,24 +172,61 @@ func (c *Controller) EnsurePlacement(cluster *myspec.M3DBCluster) error {
...
+		err := errors.New("coordinator service not found in spec")
+		c.logger.Error(err.Error())
+		c.recorder.Event(cluster, corev1.EventTypeWarning, "NoServiceConfig", "could not create m3coordinator service, missing service config in spec")
+		return err
+	}
+
+	if err := c.k8sclient.EnsureService(cluster, coordSvc); err != nil {
+		err = fmt.Errorf("error creating m3coordinator service: %v", err)
+		c.logger.Error(err.Error())
+		c.recorder.Event(cluster, corev1.EventTypeWarning, "ServiceCreateError", err.Error())
```
nerd0: I like these events, how liberal should we be with them?

pkg/controller/controller.go
```diff
@@ -125,9 +241,286 @@ func (c *Controller) Start(wg *sync.WaitGroup) error {
...
+	if !cluster.Status.HasInitializedNamespace() {
+		updated, err := c.validateNamespaceWithStatus(cluster)
+		if err != nil {
+			return err
+		}
+
+		// TODO(schallert): Decide for sure what our conventions for re-enqueueing
+		// will be.
+		if updated {
+			return errors.New("re-enqueing cluster due to api update")
```
nerd0: s/api/API/ ?

pkg/controller/controller.go
```diff
@@ -125,9 +241,286 @@ func (c *Controller) Start(wg *sync.WaitGroup) error {
...
+
+		name := fmt.Sprintf("%s-%d", cluster.Name, nextID)
+		sts := newBaseStatefulSet(name, cluster.Spec.Image, cluster.Name, zones[nextID], 1)
+
+		// Create an Affinity policy for isolationgroup.
+		nodeAffinity := &corev1.NodeAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+				NodeSelectorTerms: []corev1.NodeSelectorTerm{
+					{
+						MatchExpressions: []corev1.NodeSelectorRequirement{
```
nerd0: Can this be made local minikube friendly ? Or update the `README` to explain that a label is needed for the minikube?

```
kubectl label nodes minikube failure-domain.beta.kubernetes.io/zone=<zone_name>
```

pkg/controller/controller.go
```diff
@@ -125,9 +241,286 @@ func (c *Controller) Start(wg *sync.WaitGroup) error {
...
+		// create a statefulset
+		selfRef := metav1.NewControllerRef(cluster, schema.GroupVersionKind{
+			Group:   myspec.SchemeGroupVersion.Group,
+			Version: myspec.SchemeGroupVersion.Version,
+			// TODO(schallert): use a const here
+			Kind: "m3dbcluster",
+		})
+
+		name := fmt.Sprintf("%s-%d", cluster.Name, nextID)
+		sts := newBaseStatefulSet(name, cluster.Spec.Image, cluster.Name, zones[nextID], 1)
```
nerd0: should the `1` be replace with the `numInstances` from the spec?

cmd/m3db-operator/main.go
schallert: So far since I'm still changing stuff I've been using direct access to the clients rather than abstracting it behind something like k8sops. Once we have the full set of APIs / behaviors we depend on I think we can hide them.

cmd/m3db-operator/main.go
```diff
@@ -83,35 +94,59 @@ func main() {
...
+	go kubeInformerFactory.Start(stopCh)
+	go m3dbClusterInformerFactory.Start(stopCh)
+
 	// Init the controller
 	if err := controller.Init(); err != nil {
 		logger.Fatal("failed to init controller", zap.Error(err))
 	}
 
-	// Setup main go routines and start controller
-	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
schallert: One nice thing we get with the chan is that APIs like `WaitForCachesSync` and `wait.Until` take a chan directly and make graceful shutdown easier

pkg/controller/add_cluster.go
schallert: I definitely want to use them extensively. Even something like creating a statefulset causes 6+ info-level events. Once the behavior is more defined I think we can depend on them more

pkg/controller/update_cluster.go
```diff
@@ -63,3 +71,109 @@ func (c *Controller) updateM3DBCluster(cluster *myspec.M3DBCluster) error {
...
+	updateTime := metav1.Now()
+	cluster.Status.Conditions = append(cluster.Status.Conditions, myspec.ClusterCondition{
+		Type:           myspec.ClusterConditionPlacementInitialized,
+		Status:         corev1.ConditionTrue,
+		LastUpdateTime: updateTime.ProtoTime().String(),
+		Reason:         "PlacementCreated",
+		Message:        "Created placement",
+	})
+
+	_, err := c.crdClient.OperatorV1().M3DBClusters(cluster.Namespace).Update(cluster)
```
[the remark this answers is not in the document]
schallert: Yeah I'll add a TODO and try to abstract some of this away

pkg/controller/stateful_set.go
```diff
@@ -0,0 +1,154 @@
...
+	_configurationDirectory = "/etc/m3db/"
+	_configurationName      = "m3-configuration"
+	_configurationFileName  = "m3.yml"
+)
+
+var (
+	_configurationFileLocation = fmt.Sprintf("%s%s", _configurationDirectory, _configurationFileName)
+)
+
+func newBaseStatefulSet(ssName, image, clusterName, isolationGroup string, instanceCount int32) *appsv1.StatefulSet {
```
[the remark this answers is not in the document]
schallert: Will move

pkg/controller/controller.go
schallert: I kind of wanna keep these separated by package since there's so many imports with the same prefix. Looking through them when they're not separated can be tough to read all them. Thoughts even if it's a bit unconventional?

pkg/controller/controller.go
```diff
@@ -85,6 +193,14 @@ func New(logger *zap.Logger, kclient k8sops.K8sops) (*Controller, error) {
 	return p, nil
 }
 
+func (c *Controller) SetPlacementClient(cl placement.Client) {
```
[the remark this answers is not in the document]
schallert: Updated

pkg/controller/controller.go
schallert: Will update

pkg/controller/controller.go
schallert: Will do

pkg/controller/controller.go
schallert: Yeah I'll update the README

pkg/controller/controller.go
schallert: The more I think about it, I wonder if it just makes sense to rename this CRD to `Cluster`? I feel like M3DB is implied by the group namespace

nerd0: +1 agreed with switching to `Cluster`

schallert: I think after this PR lands we should make one to 1) switch to cluster and 2) move the api version from `v1` to `v1alpha` or `v1alpha1`

Makefile
```diff
@@ -1,10 +1,10 @@
 PROJECT_NAME := m3db-operator
 REPO_PATH    := github.com/m3db/m3db-operator
 BUILD_PATH   := $(REPO_PATH)/cmd/$(PROJECT_NAME)
-OUTPUT_DIR   := _output
-BUILD_SETTINGS := GOOS=linux GOARCH=amd64 CGO_ENABLED=0 
+OUTPUT_DIR   := out
```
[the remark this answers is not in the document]
schallert: Ah whoops I forgot I even made this change. Do we want `out` or `_output`?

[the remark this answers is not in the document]
nerd0: either one, `out` is fine.

pkg/controller/options.go
```diff
@@ -0,0 +1,96 @@
...
+	crdClient                  clientset.Interface
+	kubeClient                 kubernetes.Interface
+	kubeInformerFactory        kubeinformers.SharedInformerFactory
+	m3dbClusterInformerFactory informers.SharedInformerFactory
+	namespaceClient            namespace.Client
+	placementClient            placement.Client
+}
+
+// Option configures controller Options.
+type Option func(*Options)
```
nerd0: I am also a fan of this pattern, take a look at a [suggestion from prateek](https://github.com/m3db/m3db-operator/pull/7#discussion_r216458043). Which was implemented within other [packages](https://github.com/m3db/m3db-operator/blob/master/pkg/m3admin/namespace/namespace.go#L70-L91)

pkg/k8sops/statefulset.go
```diff
@@ -31,11 +31,18 @@ import (
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+
```
nerd0: nit: remove line.

pkg/k8sops/generators.go
```diff
@@ -253,16 +167,16 @@ func (k *k8sops) GenerateMaps(kind string, svcCfg myspec.ServiceConfiguration) m
 }
 
 // GenerateService will produce resource configured according to the spec's
-// serviceConfiguration fields
-func (k *k8sops) GenerateService(svcCfg myspec.ServiceConfiguration) *v1.Service {
+// serviceConfiguration fields.
```
nerd0: interesting, periods ?

pkg/k8sops/services.go
```diff
@@ -20,9 +20,12 @@ import (
 	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
 
 	"go.uber.org/zap"
+
 	"k8s.io/api/core/v1"
+
```
nerd0: nit: remove line.

cmd/m3db-operator/main.go
```diff
@@ -83,35 +93,70 @@ func main() {
...
+	// abstractions of statefulsets will do
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, _informerSyncDuration)
+	m3dbClusterInformerFactory := informers.NewSharedInformerFactory(crdClient, _informerSyncDuration)
+
+	opts := []controller.Option{
+		controller.WithKubeInformerFactory(kubeInformerFactory),
+		controller.WithM3DBClusterInformerFactory(m3dbClusterInformerFactory),
+		controller.WithLogger(logger),
+		controller.WithKClient(k8sclient),
+		controller.WithCRDClient(crdClient),
```
nerd0: So what was the verdict on placing the clients within the controller itself? Should we remove the k8sops interface completely? It might make testing a bit more hairy though.

Makefile
```diff
@@ -129,14 +134,14 @@ code-gen: dep-ensure ## Generate boilerplate code for kubernetes packages
 	@./hack/update-generated.sh
 
 .PHONY: build-bin
-build-bin: _output code-gen ## Build m3db-operator binary
-	which go > /dev/null || (echo "error: golang needs to be installed" && exit 1)
-	echo "building $(PROJECT_NAME)..."
+build-bin: out ## Build m3db-operator binary
```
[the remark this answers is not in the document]
schallert: Will do

README.md
```diff
@@ -2,12 +2,12 @@
```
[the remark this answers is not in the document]
schallert: Should be up-to-date now

pkg/k8sops/generators.go
schallert: I've been told that's our new convention: https://github.com/m3db/m3cluster/pull/233#discussion_r199624224

pkg/controller/update_cluster.go
```diff
@@ -21,45 +21,139 @@
...
+	})
+
+	pods, err := c.podLister.Pods(cluster.Namespace).List(labels.SelectorFromSet(targetLabels))
+	for _, pod := range pods {
+		// TODO(schallert): un-hardcode service name
+		hostname := fmt.Sprintf("%s.m3dbnode", pod.Name)
+		instance := &placementpb.Instance{
+			Id:             pod.Name,
+			IsolationGroup: pod.ObjectMeta.Labels["isolationGroup"],
+			Zone:           "embedded",
```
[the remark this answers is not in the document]
schallert: Will do

cmd/m3db-operator/main.go
schallert: I think my vote would be to wait until we have an idea of the full set of APIs we're going to need and then create an abstraction after that. Thoughts?

Re: the testing, it seems like the things that k8sops abstracts all have mock / fake packages

Makefile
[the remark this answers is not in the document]
schallert: On second thought think for code-gen we should just use the `verify-gen` target I added, and if someone hasn't ran it locally and dev'd against it CI will fail. I worry about running it any time other than local dev.

build/ci-test-minikube.sh
```diff
@@ -1,24 +1,50 @@
-#!/usr/bin/env bash 
+#!/usr/bin/env bash
 
-set -xe
+set -x
 
-# Apply the required storage for minikube 
+function with_retries() {
```
nerd0: I might be confused but where is the loop?

cmd/m3db-operator/main.go
```diff
@@ -26,32 +26,42 @@ import (
...
+	appVersion      = "0.0.1"
+	kubeCfgFile     string
+	masterHost      string
+	coordinatorAddr string
 )
 
 func init() {
 	flag.StringVar(&kubeCfgFile, "kubecfg-file", "", "Location of kubecfg file for access to kubernetes master service; --kube_master_url overrides the URL part of this; if neither this nor --kube_master_url are provided, defaults to service account tokens")
 	flag.StringVar(&masterHost, "masterhost", "http://127.0.0.1:8001", "Full url to k8s api server")
+	flag.StringVar(&coordinatorAddr, "coordinator", "", "override coordinator address if running out-of-cluster")
```
nerd0: Just a side note, in my other [PR](https://github.com/m3db/m3db-operator/pull/13) I switch from flags to environment variables. Not sure if we want to go down this route here.

cmd/m3db-operator/main.go
nerd0: That's fair since the `TODO` is there. I just hope we don't get too addicted to it and dig a hole. Works for me for now.

example/m3db-cluster.yaml
```diff
@@ -214,13 +214,15 @@ metadata:
   name: m3db-cluster
 spec:
   image: quay.io/m3/m3dbnode:latest
-  replicationFactor: 1
-  numberOfShards: 8
+  replicationFactor: 3
+  numberOfShards: 256
```
nerd0: Curious where this number came from?

hack/verify-codegen.sh
```diff
@@ -0,0 +1,48 @@
...
+
+cleanup
+
+mkdir -p "${TMP_DIFFROOT}"
+cp -a "${DIFFROOT}"/* "${TMP_DIFFROOT}"
+
+"${SCRIPT_ROOT}/hack/update-generated.sh"
+echo "diffing ${DIFFROOT} against freshly generated codegen"
+ret=0
+diff -Naupr "${DIFFROOT}" "${TMP_DIFFROOT}" || ret=$?
```
nerd0: That's hawt

build/ci-test-minikube.sh
schallert: 🤦‍♂️ I'll fix

example/m3db-cluster.yaml
schallert: My understanding is that it will speed up bootstraps since we can do more shards in parallel

PR #13, opened 2018-09-28
Adding tally prometheus reporter

nerd0:
Adding telemetry in the form of a Prometheus reporter.

### DIFF
--- a/cmd/m3db-operator/main.go
+++ b/cmd/m3db-operator/main.go
@@ -57,16 +57,16 @@ const (
 var (
 	_appVersion      = "0.0.1"
 	_kubeCfgFile     string
-	_masterHost      string
+	_masterURL       string
 	_coordinatorAddr string
 	_operatorName    = "m3db_operator"
-	_merticsPath     = "/metrics"
+	_metricsPath     = "/metrics"
 	_metricsPort     = ":8080"
 )
 
 func init() {
 	flag.StringVar(&_kubeCfgFile, "kubecfg-file", "", "Location of kubecfg file for access to kubernetes master service; --kube_master_url overrides the URL part of this; if neither this nor --kube_master_url are provided, defaults to service account tokens")
-	flag.StringVar(&_masterHost, "masterhost", "http://127.0.0.1:8001", "Full url to k8s api server")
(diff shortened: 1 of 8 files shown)


manifests/operator.yaml
```diff
@@ -83,4 +83,8 @@ spec:
           env:
             - name: OPERATOR_NAME
               value: "m3db-operator"
+            - name: ENVIRONMENT
+              value: "production" 
+            - name: KUBE_CFG_FILE
```
schallert: Do we need to specify this (even if it's empty) when we're running in-cluster?

nerd0: I figured that it would be a place holder for the human operator to know that is an option? Like documenting various configurations?

cmd/m3db-operator/main.go
```diff
@@ -43,38 +46,55 @@ import (
...
+	env := os.Getenv("ENVIRONMENT")
+	if env == "" {
+		env = "development"
+	}
+	tags := map[string]string{
+		"environment": env,
+	}
+	r := promreporter.NewReporter(promreporter.Options{})
+	scope, closer := tally.NewRootScope(tally.ScopeOptions{
+		Prefix:         os.Getenv("OPERATOR_NAME"),
```
schallert: Sounds like we can const this

cmd/m3db-operator/main.go
```diff
@@ -43,38 +46,55 @@ import (
...
+	go func() {
+		http.Handle(_metricsPath, r.HTTPHandler())
+		http.ListenAndServe(_metricsPort, nil)
+	}()
 
 	// Create k8s clients
-	crdClient, kubeClient, kubeExt, err := newKubeClient(logger, kubeCfgFile)
+	kubeCfgFile := os.Getenv("KUBE_CFG_FILE")
+	masterURL := os.Getenv("MASTER_URL")
+	crdClient, kubeClient, kubeExt, err := newKubeClient(logger, masterURL, kubeCfgFile)
```
schallert: As discussed offline let's use CLI flags here

nerd0: Done.

PR #16, opened 2018-10-01
[*] unique service names; reuse API objects

schallert:
- This reduces the API surface area of the `M3DBCluster` CRD. Instead of
  defining resource limits ourselves we can embed the API types directly
  in the cluster. This leaves flexibility for if future resource types
  are added to a cluster or if a user has custom resources they wish to
  limit.

- Similarly, this tries to reduce some boilerplate CRD config by forcing
  users to specify full services if they don't wish to have us apply the
  default ones. This leaves some more flexibility in how users can
  define services, i.e if they're on a cloud platform and want a service
  backed by a load balancer they can specify that config.

- Currently we're using the `m3dbnode` service name across all clusters,
  which will cause conflicts if we introduce multiple clusters. This
  changes the service name for the coordinator and dbnodes to include
  the name of the cluster.

- Per the API docs on StatefulSets, the headless service that controls
  pod identity for a StatefulSet must exist before the StatefulSet
  itself. This moves service creation to the beginning of the control
  loop.


pkg/controller/add_cluster.go
```diff
@@ -119,34 +111,35 @@ func (c *Controller) EnsureNamespace(cluster *myspec.M3DBCluster) (bool, error)
...
+			coordSvc,
+		)
 	}
 
-	coordSvc, found := c.getServiceConfig(_M3CoordinatorSvcName, cluster.Spec.ServiceConfigurations)
-	if !found {
-		err := errors.New("coordinator service not found in spec")
-		c.logger.Error(err.Error())
-		c.recorder.Event(cluster, corev1.EventTypeWarning, "NoServiceConfig", "could not create m3coordinator service, missing service config in spec")
+	m3dbSvc, err := k8sops.GenerateM3DBService(cluster.Name)
```
nerd0: Would this conflict if the user specifies a m3db like service within the spec? Furthermore shouldn't we allow users to pick any port configuration they wish too?

pkg/k8sops/generators_test.go
```diff
@@ -261,9 +232,13 @@ func TestGenerateStatefulSet(t *testing.T) {
...
+
 	newSS, err := GenerateStatefulSet(&fixture, isolationGroup, *instanceAmount)
 	require.Nil(t, err)
 	require.NotNil(t, newSS)
 
 	require.Equal(t, ss, newSS)
 }
+
+func TestGenerateM3DBService(t *testing.T) {
+}
```
nerd0: Missing the actual test?

PR #18, opened 2018-10-01
[k8sops] move readiness probes to new script

schallert:
As of https://github.com/m3db/m3/pull/997 we have a more accurate method
of checking whether a node is bootstrapped.

### DIFF
--- a/pkg/k8sops/generators.go
+++ b/pkg/k8sops/generators.go
@@ -50,6 +50,7 @@ const (
 	_configurationDirectory = "/etc/m3db/"
 	_configurationName      = "m3-configuration"
 	_configurationFileName  = "m3.yml"
+	_healthFileName         = "/bin/m3dbnode_bootstrapped.sh"
 
 	_labelApp             = "app"
 	_labelAppValue        = "m3db"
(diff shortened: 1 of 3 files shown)


pkg/k8sops/generators.go
```diff
@@ -142,6 +141,9 @@ func GenerateStatefulSet(
 
 	clusterSpec := cluster.Spec
 
+	// TODO(schallert): we're currently using the health of the coordinator for
+	// liveness probes until https://github.com/m3db/m3/issues/996 is fixed. Move
+	// to the dbnode's health endpoint once fixed.
```
CK-Ward: Q, do we want to health check both the coordinator and dbnode once the issue is fixed rather than replace?

schallert: Long term I think we may want to run coordinators separately from DB nodes in which case we'd health check them separately. In the short term, once that issue is fixed I think we'd have to decide which since we can only pick one port. In that case I think we'd pick the dbnode over the coordinator, since even if the coordinator hangs we could always control the cluster from another instance.

PR #27, opened 2018-10-08
[controller] handle cluster expansion

schallert:
This modifies the core event loop to support the case where a cluster
has pods which are not in the active placement.

Also included are some changes to logging such that by default the
controller runs with JSON logging, making integrations into higher-level
log viewing tools easier (such as GKE + Stackdriver logging).


schallert:
Closing and reopening (CI changes)

PR #28, opened 2018-10-08
[ci] migrate to Buildkite for CI; add checks

schallert:
This moves us to Buildkite for all operator CI needs. Additionally some
of the `make` targets were cleaned up in an attempt to make builds more
reproducible from just the `make` targets as opposed to the CI platform
having to call more steps.


Makefile
```diff
@@ -79,20 +81,46 @@ testhtml: test-internal
...
 
-.PHONY: install-mockgen
-install-mockgen: install-vendor-dep
+.PHONY: install-ci-tools
+install-ci-tools: install-vendor-dep install-codegen-tools dep-ensure install-mockgen
+	@echo "+ $@"
+	@which gocov > /dev/null || go get github.com/axw/gocov/gocov
+
+.PHONY: install-vendor-dep
+install-vendor-dep:
```
prateek: nit: can you just use retool for dep too?

Makefile
```diff
@@ -79,20 +81,46 @@ testhtml: test-internal
...
 	@echo "+ $@"
-	@which goveralls > /dev/null || (go get github.com/golang/mock/gomock && \
-		go install github.com/golang/mock/mockgen)
+	@which dep > /dev/null || (curl https://raw.githubusercontent.com/golang/dep/$(DEP_VERSION)/install.sh | DEP_RELEASE_TAG=$(DEP_VERSION) sh)
```
prateek: if you decide to use this path for dep, could you include the script in the repo or (ci-scripts). Paranoid about `curl | bash`

tools.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Tools": [
+    {
+      "Repository": "github.com/Masterminds/glide",
```
prateek: can probably remove glide

tools.json
```diff
@@ -0,0 +1,25 @@
...
+      "Commit": "5539dc14dc6c62b10c1bb1412bbdf08e7fda85bb"
+    },
+    {
+      "Repository": "github.com/m3db/build-tools/utilities/mockclean",
+      "Commit": "3e9c30b229f100027d742104ad6d6b2d968374bd"
+    },
+    {
+      "Repository": "github.com/mjibson/esc",
```
prateek: don't think you need esc/genny either

scripts/auto-gen.sh
```diff
@@ -0,0 +1,92 @@
...
+    # sed -i '' doesn't work with GNU sed, so we work around it by redirecting to a
+    # temp file first and moving it back later.
+    sed "s|$VENDOR_PATH/||" $FILE > $FILE.tmp && mv $FILE.tmp $FILE
+
+    # Strip GOPATH from the source file path
+    sed "s|Source: $GOPATH/src/\(.*\.go\)|Source: \1|" $FILE > $FILE.tmp && mv $FILE.tmp $FILE
+
+    # NB(prateek): running mockclean makes mock-gen idempotent.
+    # NB(xichen): mockclean should be run after the vendor path is stripped.
+    # NB(schallert): mockclean screws up our imports
```
prateek: lol what does it mess up?

Makefile
```diff
@@ -101,17 +129,23 @@ install-proto-bin: install-vendor-dep
...
+	@echo generating mocks
+	PATH=$(retool_bin_path):$(PATH) PACKAGE=$(package_root) $(auto_gen) $(mocks_output_dir) $(mocks_rules_dir)
 
 .PHONY: all-gen
-all-gen: mock-gen code-gen
+all-gen: mock-gen kubernetes-gen
+
+# Ensure base commit had up-to-date generated artifacts
+.PHONY: verify-all-gen
+verify-all-gen: mock-gen kubernetes-gen
```
prateek: nit: mind sticking to calling it `test-all-gen` (just to stay uniform with other repos)

Makefile
schallert: I'll go with `retool` as mentioned above

tools.json
schallert: Will do

scripts/auto-gen.sh
schallert: Without `mockclean`, `pkg/m3admin/placement/placement_mock.go` is:
```go
package placement

import (
...
placement "github.com/m3db/m3cluster/placement"
)
...
// Get mocks base method
func (m *MockClient) Get() (placement.Placement, error) {
	ret := m.ctrl.Call(m, "Get")
	ret0, _ := ret[0].(placement.Placement)
	ret1, _ := ret[1].(error)
	return ret0, ret1
}
...
```

With `mockclean`, it's
```go
package placement

import (
...
"github.com/m3db/m3cluster/placement"
)
...
// Get mocks base method
func (m *MockClient) Get() (Placement, error) {
	ret := m.ctrl.Call(m, "Get")
	ret0, _ := ret[0].(Placement)
	ret1, _ := ret[1].(error)
	return ret0, ret1
}
```

It tries to import `Placement` as if it's in this package. Guessing it's because base name of this package is the name of this package we're trying to import the type from. Didn't dig in too far past that. This is when tried with both reflect-mode mock generation and file mode.

Makefile
schallert: Will do

PR #29, opened 2018-10-09
[controller] handle cluster expansion

schallert:
This modifies the core event loop to support the case where a cluster
has pods which are not in the active placement.

Also included are some changes to logging such that by default the
controller runs with JSON logging, making integrations into higher-level
log viewing tools easier (such as GKE + Stackdriver logging).


pkg/controller/controller.go
```diff
@@ -409,9 +413,116 @@ func (c *Controller) handleClusterUpdate(key string) error {
...
+
+	// TODO(schallert): per-cluster m3admin clients
+	placement, err := c.placementClient.Get()
+	if err != nil {
+		return fmt.Errorf("error fetching active placement: %v", err)
+	}
+
+	c.logger.Info("found placement", zap.Int("currentPods", len(pods)), zap.Int("placementInsts", placement.NumInstances()))
+
+	if placement.NumInstances() < len(pods) {
```
nerd0: Curious if there should be any defense against add too many nodes? If a large amount of hosts are added in quick succession and start peer bootstrapping wouldn't that have an adverse effect on the cluster itself?

pkg/controller/controller.go
```diff
@@ -409,9 +413,116 @@ func (c *Controller) handleClusterUpdate(key string) error {
...
+		return nil
+	}
+
+	placement, err = c.placementClient.Get()
+	if err != nil {
+		return fmt.Errorf("error fetching placement: %v", err)
+	}
+
+	// See if we need to clean up the pod bootstrapping status.
+	cluster, err = c.reconcileBootstrappingStatus(cluster, placement)
```
nerd0: If pods are bootstrapping should we just keep re-queuing the cluster till it's completed?

pkg/k8sops/placement.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	setService := HeadlessServiceName(cluster.Name)
+	hostname := pod.Name + "." + setService
+
+	// TODO(schallert): dynamic zone, portname consts
+	instance := &placementpb.Instance{
+		Id:             pod.Name,
+		IsolationGroup: isoGroup,
+		Zone:           "embedded",
```
nerd0: can these be moved to `const`? Also don't we want `Zone` to be a different value per cluster?

pkg/controller/controller.go
```diff
@@ -409,9 +413,116 @@ func (c *Controller) handleClusterUpdate(key string) error {
 		}
 	}
 
+	// At this point we have the desired number of statefulsets, and every pod
```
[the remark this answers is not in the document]
schallert: Will do

pkg/controller/controller.go
schallert: I think we're protected since as soon as we try to add the second node the coordinator will throw an error since there's currently shards in not-AVAILABLE state

pkg/controller/controller.go
schallert: My plan so far was to wait until we get a new update when one of the pods transitions to ready. I've found that since the workqueue is rate-limiting, if we re-enqueue too many times it will take a really long time after that object is healthy for it to be processed depending on how many times we re-enqueue

pkg/k8sops/placement.go
schallert: Yeah will do. And I think we'll have to have zone match whatever is in the dbnode config file, so maybe we can determine it from some field in the spec

pkg/controller/controller.go
nerd0: Got it, like we discussed! Awesome that works for me.

pkg/controller/controller.go
nerd0: Got and make sense. Thanks for explaining.

pkg/k8sops/placement.go
nerd0: Yea I had a branch where I was thinking about reading the configuration map and verifying that the ports used in the configuration match the ports that were going to be used by the pods and services.  That could be expanded to match this field as well

pkg/controller/controller.go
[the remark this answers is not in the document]
schallert: As discussed offline this is coming shortly in another PR

Issue #26, opened 2018-10-04
Events


CK-Ward:
TBD

- [setup initial scaffolding for events w/EventRecord inferface] 
- [????]

PR #32, opened 2018-10-09
Create pkg for kubernetes events

CK-Ward:
Issue #26 

## Why?

- standardize kubernetes events in terms of our operator to actually be informative to the user
- Initialized events recorder package with some starter event reasons (subject to change ofc)

##  What changed?

- moved the `k8.io/client-go/tools/record` to the `util/eventer`directory
- replaced existing `c.recorder` lines with the new `eventer.Post....`lines
- removed some `c.logger` lines where i added the new `eventer` lines because the logger is passed in, this might be incorrect though so please review 

## What's next? 
- re-scope the EventReasons
- add `eventer.Post...` to wherever we see fit as we make updates to this repo 
- write a test for `eventer.go` provided it is agreed upon

### DIFF
--- a/pkg/util/eventer/eventer.go
+++ b/pkg/util/eventer/eventer.go
@@ -21,9 +21,9 @@
 package eventer
 
 import (
-	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 
+	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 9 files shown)


pkg/util/eventer/eventer.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
+)
+
+// Types of kubernetes emitted events
+const (
+
+	// Add events
+	EventReasonAdding        = "Adding"
```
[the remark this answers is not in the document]
CK-Ward: 👍 sounds good to me

CK-Ward:
> Looks good so far! I really like the events we've started to enumerate here.
> 
> In addition to the comments I left, one thing I'd like to raise is instead of each Post... function taking a recorder, what are your thoughts on having a Poster (or similarly named) interface that has functions for Normal / Warning? And then in the construction of an instance of Poster we'd pass the recorder for it to use. This would reduce boilerplate of passing the same recorder in a bunch of places, and then it would be easy to generate a mock for it that we can use in other unit tests. Additionally since we could create the Poster tied to a namespace we'd be able to send events to a specific namespace.

@schallert Yeah I thought about doing this as well, passing in `c.recorder` got annoying after a few times lol. I'll wire this up in the next pass 😎

pkg/util/eventer/eventer.go
```diff
@@ -0,0 +1,99 @@
...
+func NewEventRecorder(
+	kubeClient kubernetes.Interface,
+	logger *zap.Logger,
+	component string) record.EventRecorder {
+
+	broadcaster := record.NewBroadcaster()
+	broadcaster.StartLogging(logger.Sugar().Infof)
+	broadcaster.StartRecordingToSink(
+		&typedcorev1.EventSinkImpl{
+			Interface: kubeClient.CoreV1().Events("")})
```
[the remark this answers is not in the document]
CK-Ward: O true I'll look into it

pkg/util/eventer/eventer.go
```diff
@@ -0,0 +1,111 @@
...
+	WarningEvent(object runtime.Object, reason, message string, args ...interface{})
+}
+
+type eventer struct {
+	recorder record.EventRecorder
+}
+
+// NewEventRecorder creates a new recorder to emit kubernetes events.
+func NewEventRecorder(
+	kubeClient kubernetes.Interface,
```
nerd0: Should the constructor use option setters to keep it consistent with the other packages?

pkg/util/eventer/eventer.go
```diff
@@ -0,0 +1,111 @@
...
+func (e *eventer) NormalEvent(object runtime.Object, reason, message string) {
+	e.recorder.Event(object,
+		corev1.EventTypeNormal,
+		reason,
+		message)
+}
+
+// WarningEvent post an event of type errors or unexpectled possibly unhealthy behavior
+func (e *eventer) WarningEvent(object runtime.Object, reason, message string, args ...interface{}) {
+	e.recorder.Eventf(object,
```
schallert: Can we stick to either `Event` for both or `Eventf` for both?

pkg/util/eventer/eventer.go
schallert: Ah yeah +1

CK-Ward:
closing and reopening this

PR #43, opened 2018-10-12
[k8sops] refactor label logic into separate pkg

schallert:
Taking label logic that was in various places and consolidate in one
package.

### DIFF
--- a/pkg/controller/update_cluster.go
+++ b/pkg/controller/update_cluster.go
@@ -26,20 +26,22 @@ import (
 
 	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1"
 	"github.com/m3db/m3db-operator/pkg/k8sops"
+	"github.com/m3db/m3db-operator/pkg/k8sops/labels"
 	"github.com/m3db/m3db-operator/pkg/m3admin"
+	"github.com/m3db/m3db-operator/pkg/util/eventer"
+	"go.uber.org/zap"
 
 	"github.com/m3db/m3/src/query/generated/proto/admin"
 	"github.com/m3db/m3cluster/placement"
 
-	"github.com/m3db/m3db-operator/pkg/util/eventer"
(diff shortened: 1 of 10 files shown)


pkg/controller/update_cluster.go
```diff
@@ -26,20 +26,22 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
+	klabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
 )
 
 const (
-	_isolationGroupKey = "isolation-group"
-	_zoneEmbedded      = "embedded"
+	_zoneEmbedded = "embedded"
```
CK-Ward: just out of curiosity, what do the underscores in the front of the const mean?

schallert: I think our convention so far has been to prefix non-exported consts with an `_` to provide some extra distinction between what's a var vs. a const. I don't think it's standard Go, but I've seen it used in a few other Uber OSS repos and it's been our convention so far so just following that.

pkg/controller/update_cluster.go
```diff
@@ -209,3 +212,39 @@ func (c *Controller) reconcileBootstrappingStatus(cluster *myspec.M3DBCluster, p
...
+		err := fmt.Errorf("error adding pod to placement: %s", pod.Name)
+		c.logger.Error(err.Error())
+		return err
+	}
+
+	c.logger.Info("added pod to placement", zap.String("pod", pod.Name))
+	return nil
+}
+
+func (c *Controller) findCandidateRemovalPod(set *appsv1.StatefulSet) error {
```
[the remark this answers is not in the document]
schallert: Yeah, I messed up and ended up removing this in the next PR to satisfy the linter haha

https://github.com/m3db/m3db-operator/pull/45/files#diff-c5cea58708950416b4ffd831b961dfe4L243

PR #50, opened 2018-10-17
[controller] support instance removes

schallert:
- Support cluster instance removal w/ sanity checks.
- Implement remove operations in m3admin client.
- Define m3 placement URLs locally (remove URL has `{%s}` in it for mux
  var parsing, defining locally to "freeze" paths).

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -450,7 +450,7 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 	}
 
 	if len(unavailInsts) > 0 {
-		c.logger.Info("waiting for instances to be available", zap.Strings("instances", unavailInsts))
+		c.logger.Warn("waiting for instances to be available", zap.Strings("instances", unavailInsts))
 		return nil
 	}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -38,6 +38,8 @@ import (
 	"github.com/m3db/m3db-operator/pkg/m3admin/placement"
 	"github.com/m3db/m3db-operator/pkg/util/eventer"
 
+	m3placement "github.com/m3db/m3cluster/placement"
+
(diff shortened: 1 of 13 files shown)


pkg/controller/controller.go
```diff
@@ -440,17 +442,18 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
...
+	unavailInsts := []string{}
+	for _, inst := range placement.Instances() {
+		if !inst.IsAvailable() {
+			unavailInsts = append(unavailInsts, inst.ID())
 		}
 	}
 
+	if len(unavailInsts) > 0 {
+		c.logger.Info("waiting for instances to be available", zap.Strings("instances", unavailInsts))
+		return nil
```
nerd0: Is there any potential that non-scaling events will be dropped if the cluster receives updates for other aspects?

pkg/m3admin/placement/placement.go
```diff
@@ -39,51 +37,18 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	placementBaseURL   = "/api/v1/services/m3db/placement"
```
nerd0: Is there a `const` that can used from the `m3` repository for this ?

pkg/controller/controller.go
nerd0: The one event that I am worried about are deployments. It would require the human operator to look at the logs to understand why the deployment isn't running but that is reasonable. Could we change this to a Warn log message so that it sticks out like a sore thumb?

pkg/m3admin/placement/placement.go
nerd0: Yea this can break with either approach. I guess this is when pinning the version of M3 is going to be very important for either approach.

pkg/controller/controller.go
schallert: Yeah definitely, good point

PR #51, opened 2018-10-17
Adding helm publish script

nerd0:
Simple shell script to publish new helm charts to the s3 repository.


Makefile
```diff
@@ -186,5 +186,10 @@ build-docker: ## Build m3db-operator docker image with go binary
 	@echo "+ $@"
 	@./build/build-docker.sh
 
+.PHONY: publish-helm-charts
+publish-helm-charts: ## pushes a new version of the helm chart
+	@echo "+ $@"
+	./helm/hack/package.sh
```
schallert: Where's this script? I can't find it

nerd0: Sorry pulled a n00b mistake and didn't add it. It's now added.

helm/hack/package.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# M3 Charts 
+BUCKET=${2:-"s3://m3-helm-charts-repository/stable/"}
+CHART_NAME=${3:-"m3db-operator"}
+CHART_DIRECTORY=${4:-"$(pwd)/helm"}
+REPO_NAME="${5:-"m3-charts"}"
+VERSION=${1:-$(cat helm/${CHART_NAME}/Chart.yaml  | grep "^version" | awk '{print $2}')}
```
schallert: Could we use either env vars or getopt for these params? Or make them all mandatory?

It seems odd to have overridable defaults, but we'd only be able to override one if all priors were specified. I.e. if I want to only override `CHART_NAME`, I'd have to specify `$1` and `$2` since in order to have `$3` be a custom param there has to be a first and second param to the script

PR #52, opened 2018-10-23
Setup PersistentVolumeClaims for m3db-cluster

### LINKED ISSUE #47: Volume Claim Templates
Add to cluster definition

CK-Ward:
Closes #47 

Yay for sticky storage

# Things that need attention:

- `reclaimPolicy` in `storage-m3db-gcp.yaml` -- do we actually want Retain or should we evaluate Delete and see if it's actually ok?
- how do I look inside the PV and see that there's metric data in there...lol
- ^ based on above perhaps need to program some logic to go into the PV

### DIFF
--- a/pkg/apis/m3dboperator/v1/cluster.go
+++ b/pkg/apis/m3dboperator/v1/cluster.go
@@ -202,10 +202,10 @@ type ClusterSpec struct {
 	// +optional
 	ContainerResources corev1.ResourceRequirements `json:"containerResources" yaml:"containerResources"`
 
-	// PersistentVolumeClaims are created with predefined volumeClaimTemplates.
+	// VolumeClaimTemplates are created with predefined volumeClaimTemplates.
 	// They claim PersistentVolumes for cluster storage, volumes are dynamically
 	// provisioned by when the StorageClass is defined.
-	PersistentVolumeClaims []corev1.PersistentVolumeClaim `json:"volumeClaimTemplate" yaml:"volumeClaimTemplate"`
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplate" yaml:"volumeClaimTemplate"`
 
 	// Labels sets the base labels that will be applied to resources created by
 	// the cluster. // TODO(schallert): design doc on labeling scheme.
--- a/pkg/apis/m3dboperator/v1/cluster.go
+++ b/pkg/apis/m3dboperator/v1/cluster.go
@@ -202,9 +202,10 @@ type ClusterSpec struct {
 	// +optional
 	ContainerResources corev1.ResourceRequirements `json:"containerResources" yaml:"containerResources"`
 
-	// PersistentVolumeClaim claims a PersistentVolume for cluster storage;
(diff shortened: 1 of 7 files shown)


pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -202,6 +202,10 @@ type ClusterSpec struct {
 	// +optional
 	ContainerResources corev1.ResourceRequirements `json:"containerResources" yaml:"containerResources"`
 
+	// PersistentVolumeClaim claims a PersistentVolume for cluster storage;
+	// volumes are dynamically provisioned by when StorageClass is defined & applied
+	PersistentVolumeClaim corev1.PersistentVolumeClaim `json:"volumeClaimTemplate" yaml:"volumeClaimTemplate"`
```
schallert: Can we make this an array of claims? I could foresee a setup where we might want to have something like the commit logs on different storage than the data files since they have distinct access patterns. Also I noticed the struct field is `PersistentVolumeClaim` but the json/yaml is `volumeClaimTemplate`.

CK-Ward: Yeah I was thinking about the array of claims as well, I'll implement that.

So within the KubeAPI the stuff specified in `volumeClaimTemplates` "turns into" a `PersistentVolumeClaim` corev1 API object -- in my opinion I think it makes the most sense to still name it `PersistentVolumeClaim` because that's what actually dynamically provisions the storage whereas the templates are moreso tied to the definition of a manifest, and that's what the methods are named after (such as `corev1.PersistentVolumeCLaimLister` and what not). I can add a clarifying comment if that helps

pkg/k8sops/generators_test.go
```diff
@@ -231,6 +231,7 @@ func TestGenerateStatefulSet(t *testing.T) {
 					},
 				},
 			},
+			VolumeClaimTemplates: clusterSpec.PersistentVolumeClaims,
```
schallert: Can you add a claim template to the fixture cluster and validate that it shows up as expected here?

CK-Ward: Oh I did that but I never committed it 😅 oops

pkg/apis/m3dboperator/v1/cluster.go
```diff
@@ -202,6 +202,11 @@ type ClusterSpec struct {
 	// +optional
 	ContainerResources corev1.ResourceRequirements `json:"containerResources" yaml:"containerResources"`
 
+	// VolumeClaimTemplates are created with predefined volumeClaimTemplates.
+	// They claim PersistentVolumes for cluster storage, volumes are dynamically
+	// provisioned by when the StorageClass is defined.
+	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplate" yaml:"volumeClaimTemplate"`
```
schallert:
```suggestion
	VolumeClaimTemplates []corev1.PersistentVolumeClaim `json:"volumeClaimTemplates" yaml:"volumeClaimTemplates"`
```

schallert: One typo re: plurality

PR #54, opened 2018-10-26
[api] support custom namespaces in spec

### LINKED ISSUE #41: Allow custom namespace options in cluster CRD
We currently set opinionated defaults for cluster namespaces. We should allow customization of namespace + retention options.

schallert:
Allows specifying either full namespace options or pointing to a preset.

Closes #41.


pkg/apis/m3dboperator/v1/namespace.go
```diff
@@ -0,0 +1,95 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package v1
+
+import "time"
+
+// Namespace defines a namespace or points to a preset namespace.
```
CK-Ward: I noticed on line 68 `NamespaceOptions` has a comment saying it's an M3DB namespace, do we want to explicitly say in this comment how this is a k8 namespace?

pkg/controller/controller.go
```diff
@@ -396,20 +396,14 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
...
-		// TODO(schallert): Decide for sure what our conventions for re-enqueueing
-		// will be. EDIT: just return the cluster object from funcs and continue
-		// using that version of it.
-		if updated {
-			err = errors.New("re-enqueing cluster due to API update")
-			c.recorder.WarningEvent(cluster, eventer.ReasonFailedToUpdate, err.Error())
-		}
+	if len(cluster.Spec.Namespaces) == 0 {
+		c.logger.Warn("cluster has no namespaces defined", zap.String("cluster", cluster.Name))
+		c.recorder.WarningEvent(cluster, eventer.ReasonUnknown, "cluster %s has no namespaces", cluster.Name)
```
CK-Ward: So the recorder has the logger passed in at time of construction, do you know if is it necessary to both log and event? I was going to investigate this more thoroughly but I haven't really gotten to it

pkg/m3admin/namespace/client.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Delete will delete a namespace
+func (n *namespaceClient) Delete(namespace string) error {
+	url := fmt.Sprintf("%s%s/%s", n.url, nsh.AddURL, namespace)
+	_, err := n.client.DoHTTPRequest("DELETE", url, nil)
+	if err != nil {
+		return err
+	}
+	n.logger.Info("successfully deleted namespace")
```
CK-Ward: Maybe an event here? Or we can just call events wherever this is called for some more contextual events.

pkg/m3admin/namespace/namespace.go
```diff
@@ -21,179 +21,78 @@
...
-			return err
-		}
-		n.url = u
-		return nil
-	})
-}
+// RequestFromSpec returns a namespace add request from a cluster spec namespace
+// config.
+func RequestFromSpec(ns myspec.Namespace) (*admin.NamespaceAddRequest, error) {
+	if ns.Name == "" {
```
CK-Ward: Is this a generally accepted way in golang to check for an empty string? I usually go with `'len(string) <= 0` but maybe that's just a java thing

pkg/controller/controller.go
CK-Ward: Ah okay I'm not super familiar with zap and didn't realize it wasn't just strings lol this is fine

pkg/controller/add_cluster.go
```diff
@@ -81,29 +80,6 @@ func (c *Controller) EnsurePlacement(cluster *myspec.M3DBCluster) error {
 	return nil
 }
 
-// EnsureNamespace will retrieve current namespaces to ensure one matches the
-// cluster name or create a new namespace to match the cluster name. Returns a
-// bool indicating whether a namespace was created and any errors encountered.
-func (c *Controller) EnsureNamespace(cluster *myspec.M3DBCluster) (bool, error) {
```
CK-Ward: So just to get my understanding correct, we're removing this function because the namespace is already "ensured" in the yaml with its own customizations?

pkg/controller/common_test.go
```diff
@@ -80,6 +81,8 @@ func (deps *testDeps) newController() *Controller {
 
 		placementClient: deps.placementClient,
 		namespaceClient: deps.namespaceClient,
+
+		recorder: eventer.NewNopPoster(),
```
CK-Ward: 🎉

pkg/controller/update_cluster.go
```diff
@@ -45,49 +47,111 @@ import (
...
 	}
 
-	// We updated an API object so tell the queue to re-enqueue.
-	return true, nil
+	return nil
+}
+
+// cleanupNamespaces will delete any namespaces in the m3db cluster that aren't
+// in the spec.
+func (c *Controller) cleanupNamespaces(cluster *myspec.M3DBCluster, registry *dbns.Registry) error {
```
CK-Ward: I'm not sure `cleanup` is the most effective name here since it's just deleting namespaces and not organizing them or anything else that can be related to cleaning. Maybe `prune` is a better word?

pkg/controller/add_cluster.go
schallert: Ah yeah, I removed it because I thought that 1) now that we have multiple namespaces just `EnsureNamespace` didn't fit that description, and 2) since we're more than just ensuring it's there we might end up removing some, so "reconcile" felt more fitting. We still have to take what's in the YAML and "ensure" they end up in the cluster, but it's a bit more of a dynamic process

pkg/controller/update_cluster.go
schallert: I like that much better! I'll update

PR #55, opened 2018-10-31
[controller] support per-cluster m3admin clients

schallert:
Adds a multi-admin client that multiplexes over many m3admin clients.


pkg/controller/options.go
```diff
@@ -108,25 +105,16 @@ func WithM3DBClusterInformerFactory(f informers.SharedInformerFactory) Option {
...
-		o.namespaceClient = cl
-	})
-}
-
-// WithPlacementClient sets the m3db placement client. If unset a default
-// in-cluster client will be created.
-func WithPlacementClient(cl placement.Client) Option {
-	return optionFn(func(o *options) {
-		o.placementClient = cl
+		o.kubectlProxy = true
```
CK-Ward: wouldn't this be `o.kubectlProxy = use`?

PR #56, opened 2018-11-13
[m3admin] Add Replace to placement client

schallert:


pkg/m3admin/placement/types.go
```diff
@@ -38,4 +38,6 @@ type Client interface {
 	Add(instance placementpb.Instance) error
 	// Remove removes a given instance with the given ID from the placement.
 	Remove(id string) error
+	// Replace replaces one instance with another.
+	Replace(leavingID string, newInstance placementpb.Instance) error
```
CK-Ward: i don't know how much i like the name `leavingID` but i don't have any better ideas lol

pkg/controller/m3admin_client.go
```diff
@@ -221,3 +221,7 @@ func (c errorPlacementClient) Add(instance placementpb.Instance) error {
 func (c errorPlacementClient) Remove(id string) error {
 	return c.err
 }
+
+func (c errorPlacementClient) Replace(_ string, _ placementpb.Instance) error {
```
CK-Ward: why do these parameters have underscores?

pkg/m3admin/placement/types.go
schallert: Ah ya know what, I'm going to change it to `leavingInstanceID` to match the placement service [API](https://godoc.org/github.com/m3db/m3cluster/placement#Service) and the [proto](https://github.com/m3db/m3/blob/85841a3ff2b11b69549e6ae2f19cfa750408c850/src/query/generated/proto/admin/placement.proto#L25)

pkg/controller/m3admin_client.go
schallert: It's a thing I've seen in Go before to indicate the parameter isn't used in the function, but I just realized I can drop the `_` entirely! Going to do that

PR #63, opened 2018-11-26
[api] make pod identity configurable

schallert:
We need a manner of determining pod uniqueness, as pod names remain the
same across restarts, rescheduling of storage, etc. This PR introduces
the concept of "pod identity", which is created from a configurable list
of sources to determine the uniqueness of a pod.

This adds a new controller that watches pods, and if they're part of an
M3DB cluster annotates them with their identity. This annotation is
passed to the pod through a downward API volume and picked up by dbnode
instances when they start up and look for themselves in the placement.
All placement operations are based off of this identity.


pkg/podidentitycontroller/controller.go
```diff
@@ -0,0 +1,337 @@
...
+	controllerName = "m3db-podidentity-controller"
+	workQueueName  = "m3db-pod-identity"
+)
+
+var (
+	errOrphanedPod = errors.New("pod does not belong to an m3db cluster")
+)
+
+// Controller object.
+type Controller struct {
```
CK-Ward: What did we decide we were going to do about this extra controller for now? Whatever it is, in the meantime can we make an issue to make the Controller struct reusable so our repo can be DRY lol

schallert: Yeah you're spot on here, this is a significantly better idea 😂 . Ripped out a ton of the boiler plate in exchange for 2 work queues in the main controller, we can always split out later if it gets crowded in that code and we invest in better controller construction.

pkg/controller/controller.go
```diff
@@ -93,8 +97,9 @@ type Controller struct {
 	podLister          corelisters.PodLister
 	podsSynced         cache.InformerSynced
 
-	workQueue workqueue.RateLimitingInterface
-	recorder  eventer.Poster
+	clusterWorkQueue workqueue.RateLimitingInterface
+	podWorkQueue     workqueue.RateLimitingInterface
```
CK-Ward: yay this structure makes a lot more sense lolol

PR #65, opened 2018-12-05
[api] allow customizing configmap; add default

### LINKED ISSUE #60: Support user-specified configmaps + hardcoded default
We currently hardcode the ConfigMap name, as we have a sane base config. We should allow the ConfigMap name the be specified for more complex configurations, as well as consider hardcoding the "blessed" one in the binary so users don't have to create it.

schallert:
- Allows customizing of the configmap used by the cluster rather than
  hardcoding.
- Adds a default configmap embedded with the binary.
- Remove custom services from being embedded in the API.
- Default to EmptyDir if no PVC templates specified.
- Add node-level fields to pod ID sources.
- Add `1m:40d` preset.
- Always include pod name in ID (we depend on this for replaces)

Closes #60


pkg/apis/m3dboperator/v1/pod_identity.go
```diff
@@ -26,20 +26,36 @@ type PodIdentitySource string
...
-	PodIdentitySourcePodName PodIdentitySource = "PodName"
+
+	// PodIdentitySourceNodeSpecExternalID derives identity from the 'externalID'
+	// spec field of the node. Note this field was deprecated after Kubernetes
+	// 1.11.
+	PodIdentitySourceNodeSpecExternalID PodIdentitySource = "NodeSpecExternalID"
+
+	// PodIdentitySourceNodeSpecProviderID derives identity from the 'providerID'
+	// spec field of the node.
+	PodIdentitySourceNodeSpecProviderID PodIdentitySource = "NodeSpecProviderID"
```
CK-Ward: what is the difference between External ID and Provider ID? Like, what do they actually mean in this context

schallert: It's kind of weird, but provider ID is the node name but external ID is the unique underlying VM ID, iirc it's the `id` field in these docs: https://cloud.google.com/compute/docs/storing-retrieving-metadata#default

I think we have to clean this up another PR since the field becomes deprecated in 1.11

PR #67, opened 2018-12-08
Replace m3db instances

### LINKED ISSUE #58: Replace clusters
Have the controller replace clusters w/correct storage attached when necessary

CK-Ward:
Closes #58

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -449,7 +449,19 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 	}
 
 	// check if any pods inside the cluster need to be swapped in
-	c.checkPodsForReplacement(cluster, pods, placement)
+	leavingInstanceID, podToReplace, err := c.checkPodsForReplacement(cluster, pods, placement)
+	if err != nil {
+		return err
+	}
+
+	if podToReplace != nil {
+		err = c.replacePodInPlacement(cluster, placement, leavingInstanceID, podToReplace)
+		if err != nil {
+			c.recorder.WarningEvent(cluster, eventer.ReasonFailedToUpdate, "could not replace instance: " + leavingInstanceID)
+			return err
+		}
+		c.recorder.NormalEvent(cluster, eventer.ReasonSuccessfulUpdate, "successfully replaced instance: " + leavingInstanceID)
+	}
 
 	for _, set := range childrenSets {
 		zone, ok := set.Labels[labels.IsolationGroup]
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -328,7 +328,7 @@ func (c *Controller) handleClusterEvent(key string) error {
 // We are guaranteed by handleClusterEvent that we will never be passed a nil
 // cluster here.
 func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
-	// MUST create a deep copy of the cluster or risk corruping cache! Technically
+	// MUST create a deep copy of the cluster or risk corrupting cache! Technically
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -446,6 +448,9 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 		return err
 	}
 
+	// check if any pods inside the cluster need to be swapped in
+	c.checkPodsForReplacement(cluster, pods, placement)
```
schallert: Do we need to call `replacePodInPlacement` on the result we get here?

pkg/controller/update_cluster.go
```diff
@@ -308,6 +309,78 @@ func (c *Controller) addPodToPlacement(cluster *myspec.M3DBCluster, pod *corev1.
 	return nil
 }
 
+func (c *Controller) checkPodsForReplacement(
+	cluster *myspec.M3DBCluster,
+	pods []*corev1.Pod,
+	pl placement.Placement) (string, *corev1.Pod, error) {
+
+	var instancePodID myspec.PodIdentity
```
schallert: Can we move this declaration to inside the `for` loop? I'm not sure what happens if on one iteration it has fields A and B, and the next object only deserialized field A, if B from the first loop would be left hanging

pkg/controller/update_cluster.go
```diff
@@ -308,6 +309,78 @@ func (c *Controller) addPodToPlacement(cluster *myspec.M3DBCluster, pod *corev1.
...
+			return "", nil, err
+		}
+
+		for _, inst := range insts {
+			if strings.EqualFold(strings.Split(inst.Hostname(), ".")[0], pod.pod.Name) {
+				if err = json.Unmarshal([]byte(inst.ID()), &instancePodID); err != nil {
+					return "", nil, err
+				}
+
+				if !strings.EqualFold(clusterPodID.UID, instancePodID.UID) {
```
schallert: Since I think we're going to assume we always have `Name` but might not have `UID` based on user config, can we check `Name` here instead?

pkg/controller/update_cluster.go
```diff
@@ -308,6 +309,78 @@ func (c *Controller) addPodToPlacement(cluster *myspec.M3DBCluster, pod *corev1.
...
+
+	for _, pod := range sortedPods {
+
+		clusterPodID, err := c.podIDProvider.Identity(pod.pod, cluster)
+		if err != nil {
+			return "", nil, err
+		}
+
+		for _, inst := range insts {
+			if strings.EqualFold(strings.Split(inst.Hostname(), ".")[0], pod.pod.Name) {
```
schallert: I like that this is more efficient and that we can skip unmarshalling unnecessarily based on this check! I wonder if it's worth skipping this check and just using the unmarshalled check below in case we ever want to use IPs for hostnames or something

pkg/controller/controller.go
CK-Ward: yes oops my bad lol

pkg/controller/update_cluster.go
CK-Ward: yeah that makes more sense 👍

pkg/controller/update_cluster.go
CK-Ward: if we check only `Name` though how do we know the pod is a valid case for replacement?

pkg/controller/update_cluster.go
schallert: I don't think we should block merging on this since we know we only use hostname, but can we create an issue to follow up?

CK-Ward: yeah definitely 👍

PR #70, opened 2018-12-12
[api] support Kubernetes 1.11

schallert:
Pulls in more recent client versions and updates usage of a
to-be-deprecated field.

Tested by running against a 1.10 and 1.11 cluster.

Had to make some changes to ensure the same status update functions work
across both versions and didn't trigger constant loop updates. Also
modified the re-enqueue logic to refresh our view of the world rather
than re-entering the loop.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -405,14 +405,8 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 	}
 
 	if !cluster.Status.HasInitializedPlacement() {
-		updated, err := c.validatePlacementWithStatus(cluster)
+		cluster, err = c.validatePlacementWithStatus(cluster)
 		if err != nil {
-			return err
-		}
-
-		if updated {
-			err = errors.New("re-enqueing cluster due to API update")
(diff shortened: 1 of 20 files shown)


pkg/controller/controller.go
```diff
@@ -463,10 +457,10 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 	if podToReplace != nil {
 		err = c.replacePodInPlacement(cluster, placement, leavingInstanceID, podToReplace)
 		if err != nil {
-			c.recorder.WarningEvent(cluster, eventer.ReasonFailedToUpdate, "could not replace instance: " + leavingInstanceID)
+			c.recorder.WarningEvent(cluster, eventer.ReasonFailedToUpdate, "could not replace instance: "+leavingInstanceID)
 			return err
 		}
-		c.recorder.NormalEvent(cluster, eventer.ReasonSuccessfulUpdate, "successfully replaced instance: " + leavingInstanceID)
+		c.recorder.NormalEvent(cluster, eventer.ReasonSuccessfulUpdate, "successfully replaced instance: "+leavingInstanceID)
```
schallert: Dunno why this updated but some auto-formatter

PR #74, opened 2018-12-13
[readme] Remove more outdated content; more reqs

schallert:


schallert:
Going to pull this in with #66

PR #81, opened 2019-02-11
Minor code cleanup of m3db_operator

tariq1890:
- Run gofmt simplify on generators_test.go, statefulset.go, and update_cluster_test.go
 - Replace Version(deprecated) with Versions in GenerateCRD
 - Fix superfluous err!=nil check in update_cluster.go

These changes should be harmless and non-breaking :).


tariq1890:
Requesting a review on this :)

PR #83, opened 2019-02-20
[config] add tagOptions config for coordinator

schallert:
This required config parameter was recently added to the coordinator and
needs to be updated here.


dileepranganathan:
@schallert Is the tagOptions config required for the example config-map as well ??[https://github.com/m3db/m3db-operator/blob/master/example/config-map.yaml](url)

schallert:
> @schallert Is the tagOptions config required for the example config-map as well ??https://github.com/m3db/m3db-operator/blob/master/example/config-map.yaml

It is! Thanks for the callout @dileepranganathan :) . Putting up a PR

PR #94, opened 2019-02-28
[api] represent durations as strings

schallert:
We'd previously been using `time.Duration` as the type of various
durations in the namespace options. I swear this was working a while ago
so I'm not sure if something changes with `client-go` or if I totally
missed and issue. This changes the CRD to use strings and parse the
durations separately, which should fix the issue of the string
durations failing to parse.

Also added some context to the namespace create error and include it in
an event.

### DIFF
--- a/pkg/m3admin/namespace/namespace.go
+++ b/pkg/m3admin/namespace/namespace.go
@@ -23,6 +23,7 @@ package namespace
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1alpha1"
 
@@ -46,9 +47,14 @@ func RequestFromSpec(ns myspec.Namespace) (*admin.NamespaceAddRequest, error) {
 	}
 
 	if ns.Options != nil {
+		opts, err := requestOptsFromAPI(ns.Options)
+		if err != nil {
+			return nil, err
+		}
+
 		return &admin.NamespaceAddRequest{
 			Name:    ns.Name,
-			Options: requestOptsFromAPI(ns.Options),
+			Options: opts,
 		}, nil
 	}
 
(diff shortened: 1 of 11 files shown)


schallert:
cc @jchanam this should fix the long term issue you were running into!

pkg/m3admin/namespace/namespace.go
```diff
@@ -62,39 +68,84 @@ func RequestFromSpec(ns myspec.Namespace) (*admin.NamespaceAddRequest, error) {
...
+		RetentionOptions:  retentionOpts,
 		SnapshotEnabled:   opts.SnapshotEnabled,
-		IndexOptions:      indexOptsFromAPI(opts.IndexOptions),
-	}
+		IndexOptions:      indexOpts,
+	}, nil
 }
 
-func retentionOptsFromAPI(opts myspec.RetentionOptions) *m3ns.RetentionOptions {
+func retentionOptsFromAPI(opts myspec.RetentionOptions) (*m3ns.RetentionOptions, error) {
```
nerd0: nit: should this be validateRetentionOpts?

pkg/m3admin/namespace/namespace.go
```diff
@@ -62,39 +68,84 @@ func RequestFromSpec(ns myspec.Namespace) (*admin.NamespaceAddRequest, error) {
...
+		BufferPastNanos:                          bufferPast.Nanoseconds(),
 		BlockDataExpiry:                          opts.BlockDataExpiry,
-		BlockDataExpiryAfterNotAccessPeriodNanos: opts.BlockDataExpiryAfterNotAccessPeriod.Nanoseconds(),
-	}
+		BlockDataExpiryAfterNotAccessPeriodNanos: expiryNanos.Nanoseconds(),
+	}, nil
 }
 
-func indexOptsFromAPI(opts myspec.IndexOptions) *m3ns.IndexOptions {
+func indexOptsFromAPI(opts myspec.IndexOptions) (*m3ns.IndexOptions, error) {
```
nerd0: ditto with rename.

pkg/m3admin/namespace/namespace.go
schallert: Hmm I don't quite feel that it's validating, because it's taking one representation (our API) of retention options and transforming it into a new representation (the M3DB API). Maybe `transformRetentionOpts`? But then it's not super clear what's being transformed. Curious your thoughts

nerd0: yea this is very nitty, when reading the invocation above I wasn't sure what this function did. The reason I mentioned validating is due to the transforming of the values to ensure that they are parsed correctly i.e. the error handling. I think `transformRetentionOpts` is not bad or perhaps `normalizeRetentionOpts`? again super nitty and not a big deal in the end.

schallert: Yeah good call, updated

PR #98, opened 2019-03-01
[api] support per-isogroup storageclasses

### LINKED ISSUE #92: Mismatch between availability zones of pods and pv
* What version of Kubernetes are you running? Please include the output of `kubectl version`.

v1.11.6

* What are you trying to do?

Creating a m3db cluster with persistent storage in AWS with EBS

* What did you expect to happen?

The pods to start with a pvc in the same availability zone

* What happened?

After applying the following m3db spec:
```
apiVersion: operator.m3db.io/v1alpha1
kind: M3DBCluster
metadata:
  name: m3db-cluster
  namespace: m3db
spec:
  image: quay.io/m3db/m3dbnode:latest
  replicationFactor: 3
  numberOfShards: 256
  isolationGroups:
    - name: eu-west-1b
      numInstances: 1
    - name: eu-west-1c
      numInstances: 1
    - name: eu-west-1a
      numInstances: 1
  namespaces:
    - name: short
      preset: 10s:2d
    - name: medium
      preset: 1m:40d
  configMapName: m3db-cluster-m3-configuration
  containerResources:
    requests:
      memory: 16Gi
      cpu: "1"
    limits:
      memory: 18Gi
      cpu: "4"
  dataDirVolumeClaimTemplate:
    metadata:
      name: m3db-cluster-data
    spec:
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          storage: 500Gi
        limits:
          storage: 500Gi
```

The pods are not being able to schedule because the pod is requested to be in the AZ `eu-west-1a`:

```
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: failure-domain.beta.kubernetes.io/zone
            operator: In
            values:
            - eu-west-1a
```

But the PV was created in another one:

```
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    kubernetes.io/createdby: aws-ebs-dynamic-provisioner
    pv.kubernetes.io/bound-by-controller: "yes"
    pv.kubernetes.io/provisioned-by: kubernetes.io/aws-ebs
  labels:
    failure-domain.beta.kubernetes.io/region: eu-west-1
    failure-domain.beta.kubernetes.io/zone: eu-west-1b
  name: pvc-25500b63-3aae-11e9-b9a5-0a16e40a2c44
```

Because the claim is general for all the statefulsets created, there's no way of setting a different storage class for each zone.

Do you know how to bypass this error?

schallert:
Users running Kubernetes < 1.12, in which topology-aware volume
scheduling was introduced, cannot use a single storageclass for all
their statefulsets as there is no guarantee that the PV for a pod will
be created in the zone the pod is pinned to. This can cause pods to be
stuck in an unschedulable state.

This allows users to specify a StorageClass per-statefulset (zone). This
solves the above problem as a single StorageClass can be created with
attributes to specify that PVs created with it will be in a given zone
(see
https://v1-11.docs.kubernetes.io/docs/concepts/storage/storage-classes/).

Fixes #92

### DIFF
--- a/example/m3db-cluster-per-zone-storageclass.yaml
+++ b/example/m3db-cluster-per-zone-storageclass.yaml
@@ -33,7 +33,7 @@ metadata:
 spec:
   image: quay.io/m3db/m3dbnode:latest
   replicationFactor: 3
-  numberOfShards: 4
+  numberOfShards: 1024
   isolationGroups:
   - name: us-east1-b
     numInstances: 1
--- a/example/m3db-cluster-per-zone-storageclass.yaml
+++ b/example/m3db-cluster-per-zone-storageclass.yaml
@@ -1,0 +1,64 @@
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: fast-east1-b
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-ssd
+  zone: us-east1-b
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: fast-east1-c
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-ssd
+  zone: us-east1-c
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: fast-east1-d
+provisioner: kubernetes.io/gce-pd
+parameters:
+  type: pd-ssd
+  zone: us-east1-d
+---
+apiVersion: operator.m3db.io/v1alpha1
+kind: M3DBCluster
+metadata:
+  name: m3db-cluster-pv
+spec:
+  image: quay.io/m3db/m3dbnode:latest
+  replicationFactor: 3
+  numberOfShards: 4
+  isolationGroups:
+  - name: us-east1-b
+    numInstances: 1
+    storageClassName: fast-east1-b
+  - name: us-east1-c
+    numInstances: 1
+    storageClassName: fast-east1-c
+  - name: us-east1-d
+    numInstances: 1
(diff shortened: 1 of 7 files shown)


example/m3db-cluster-per-zone-storageclass.yaml
```diff
@@ -0,0 +1,64 @@
...
+  zone: us-east1-d
+---
+apiVersion: operator.m3db.io/v1alpha1
+kind: M3DBCluster
+metadata:
+  name: m3db-cluster-pv
+spec:
+  image: quay.io/m3db/m3dbnode:latest
+  replicationFactor: 3
+  numberOfShards: 4
```
schallert: Note to self: change to 1024 or larger

pkg/k8sops/generators.go
```diff
@@ -188,6 +192,9 @@ func GenerateStatefulSet(
 	} else {
 		template := cluster.Spec.DataDirVolumeClaimTemplate.DeepCopy()
 		template.ObjectMeta.Name = _dataVolumeName
+		if sc := isolationGroup.StorageClassName; sc != "" {
```
nerd0: Is there anyway to ensure that there is a storage class of that name within the template itself? Do that need to validated?

schallert: I think for now it's better to let Kube throw the error when it tries to create the statefulset and doesn't find it, otherwise we'd have to add new RBAC permissions to read StorageClasses

nerd0: Fair, I wanted to ensure that it would be clear to user that a storage class doesn't exist.

PR #99, opened 2019-03-06
[api] cluster spec etcd endpts; per-cluster envs

### LINKED ISSUE #64: Allow specifying of etcd clusters in CRD
The etcd clusters m3db reads its placement from is one of the only dynamic parts of the configs we may expect users to customize or change frequently. By including the etcd clusters as part of the CRD and templating them out for m3db's config, we could eliminate most use cases where users would need to specify custom cluster config maps (most of the other settings are sane defaults).

### LINKED ISSUE #88: data for multiple clusters can conflict in etcd
Under default settings, the placement keys for multiple clusters in etcd may conflict. We should ensure that the environments under which we store cluster metadata in etcd are unique per-m3db cluster.

schallert:
This PR has a few changes to improve the interaction between
operator-operated M3DB clusters and etcd:

1. Users can now specify etcd endpoints on the m3dbcluster spec,
removing one of the most common changes users would have to make in
their config files.

2. The operator will include a per-cluster environment header in all
requests to the coordinator, meaning that namespaces and placements will
be stored under separate keys per-cluster in etcd and multiple
m3dbclusters can share the same etcd cluster.

**NOTE**: This PR introduces a breaking change that will require users
to 1) rename 2 keys in their etcd clusters and 2) change a value in
their configmap if they would like the operator to continue being able
to manage clusters from before this commit. The changelog for this
release will include instructions for this.

Fixes #64 
Fixes #88


cmd/m3db-operator/main.go
```diff
@@ -53,7 +53,7 @@ import (
 
 const (
 	// Informers will resync on this interval
-	_informerSyncDuration = 30 * time.Second
+	_informerSyncDuration = 5 * time.Minute
```
nerd0: Why the increase?

docs/api.md
```diff
@@ -43,7 +43,8 @@ ClusterSpec defines the desired state for a M3 cluster to be converge to.
 | numberOfShards | NumberOfShards defines how many shards in total | int32 | false |
 | isolationGroups | IsolationGroups specifies a map of key-value pairs. Defines which isolation groups to deploy persistent volumes for data nodes | [][IsolationGroup](#isolationgroup) | false |
 | namespaces | Namespaces specifies the namespaces this cluster will hold. | [][Namespace](#namespace) | false |
-| configMapName | ConfigMapName specifies the ConfigMap to use for this cluster. If unset a sane default will be used. | *string | false |
+| etcdEndpoints | EtcdEndpoints defines the etcd endpoints to use for service discovery. Must be set if no custom configmap is defined. If set, etcd endpoints will be templated in to the default configmap template. | []string | false |
+| configMapName | ConfigMapName specifies the ConfigMap to use for this cluster. If unset a default configmap with template variables for etcd endpoints will be used. | *string | false |
```
nerd0: should we link the default template that will be used as `default-config` above?

pkg/controller/m3admin_client_test.go
```diff
@@ -87,17 +88,18 @@ func TestNewMultiAdminClient(t *testing.T) {
...
-		m.adminClient,
+		m.adminClientFn,
 		m.logger,
 	} {
 		assert.NotNil(t, v)
 	}
 }
 
-func newM3DBCluster(name string) *myspec.M3DBCluster {
+func newM3DBCluster(ns, name string) *myspec.M3DBCluster {
```
nerd0: Is this assuming a 1:1 association of namespace to cluster?

pkg/controller/m3admin_client.go
```diff
@@ -76,19 +81,26 @@ func clusterURLProxy(cluster *myspec.M3DBCluster) string {
 	return url
 }
 
-func newMultiAdminClient(m3adminClient m3admin.Client, logger *zap.Logger) *multiAdminClient {
+func newMultiAdminClient(adminOpts []m3admin.Option, logger *zap.Logger) *multiAdminClient {
```
nerd0: should the logger just be another another `m3admin.Option`?

cmd/m3db-operator/main.go
schallert: It made the logs super spammy, and we do a good job of reacting to events so I didn't feel it was necessary to have such a frequent resync. I browsed some other operators code and it seems like they do 5m-30m

docs/api.md
schallert: Ah yeah good point. I think I'll add a comment saying "see configuration section of docs" just so we don't have too many links in the comments

pkg/controller/m3admin_client_test.go
schallert: I don't think so, but it's just generating a fake metadata for the tests

pkg/controller/m3admin_client.go
schallert: Actually I just realized we need it because we pass the option to the namespace/placement clients which take a different `Option` type than `m3admin.Option`

nerd0: Well that's annoying

PR #106, opened 2019-03-20
[api] support custom node affinity per-isogroup

### LINKED ISSUE #68: Enable aliasing of isolation groups
We currently define isolation groups as the name of a zone and the number of desired instances in that zone. For example:
```
isolationGroups:
- name: us-east1-b
  numInstances: 1
- name: us-east1-c
  numInstances: 1
- name: us-east1-d
  numInstances: 1
```

The `Name` field maps directly to 1) the `isolationGroup` field in the M3DB placement, and 2) the value of the `failure-domain.beta.kubernetes.io/zone` label on a node. This means that one cannot create clusters with RF=3 within a single.

This has been suitable for us so far as our focus has been on creating highly available regional M3DB clusters with strict zonal isolation. However there may be use cases where users want to have RF=3 within a single zone, and we should accommodate that.

One possible solution would be to alias isolation groups, wherein multiple `isolationGroups` in the M3DB placement could map to separate zones in the Kubernetes cluster. For example:

```
isolationGroups:
- name: isoGroup1
  zoneValue: us-east1-b
  numInstances: 1
- name: isoGroup2
  zoneValue: us-east1-b
  numInstances: 1
- name: isoGroup3
  zoneValue: us-east1-b
  numInstances: 1
```

In this case we'd have an RF=3 cluster within `us-east1-b`, but 3 unique isolation groups for M3DB to replicate shards across.

schallert:
Previously we had been using the name of an IsolationGroup as its
the value selector in a StatefulSet's NodeAffinity, with the key
hardcoded to `failure-domain.beta.kubernetes.io/zone`. This meant that
users had to have clusters across 3 unique zones.

This relaxes those constraints and allows users to specify the
NodeAffinity selectors they wish to use, meaning they can have clusters
within a single zone or other failure domain.

Fixes #68


cmd/m3db-operator/main.go
```diff
@@ -98,10 +97,6 @@ func main() {
 
 	defer logger.Sync()
 
-	logger.Info("Go", zap.Any("VERSION", runtime.Version()))
```
nerd0: Why get ride of this?

pkg/controller/add_cluster.go
```diff
@@ -26,67 +26,16 @@ import (
...
-	// defaults for placement init request
-	_defaultM3DBPort = 9000
-	_defaultZone     = "embedded"
 )
 
 var (
 	errEmptyConfigMap = errors.New("ConfigMapName cannot be empty if non-nil")
 )
 
-// EnsurePlacement ensures that a placement exists otherwise create one
```
nerd0: I assume this code was replaced beforehand, didn't see it replaced within this PR.

scripts/run_e2e_tests.sh
```diff
@@ -46,7 +46,7 @@ function main() {
 
   trap cleanup EXIT
 
-  kubectl proxy &
+  kubectl proxy &>/dev/null &
```
nerd0: Why silence the output, couldn't it help if the test borks?

cmd/m3db-operator/main.go
schallert: Since I added the m3x build reporter we get all of this info for free in a startup log from there

pkg/controller/add_cluster.go
schallert: Yeah turns out I couldn't find it used anywhere

scripts/run_e2e_tests.sh
schallert: So the annoying this was this would interleave the proxy output with the test and it was confusing what was coming from where, but I guess that's better than nothing so will re-add

cmd/m3db-operator/main.go
nerd0: nice.

scripts/run_e2e_tests.sh
baxor: You could consider something like: 
`kubectl proxy |sed -e 's/^/kube proxy: /'&`

PR #112, opened 2019-04-03
[config] set new default m3db bootstrapper config

schallert:
M3DB recently changed the default configuration to add
`returnUnfulfilledForCorruptCommitLogFiles: false` by default. This
makes it easier to recover from catastrophic cluster failures, so we'll
set this by default.

More from the M3DB docs: https://docs.m3db.io/operational_guide/availability_consistency_durability/


schallert:
@richardartoul if you have any thoughts on other defaults we should update would love to hear them!

PR #114, opened 2019-04-04
[ci] build + release w/ go 1.12

schallert:


Makefile
```diff
@@ -44,9 +45,11 @@ metalint_check       := .ci/metalint.sh
 metalint_config      := .metalinter.json
 metalint_exclude     := .excludemetalint
 gopath_prefix        := $(GOPATH)/src
+gopath_bin_path      := $(GOPATH)/bin
```
mway: note: i'd probably do something like this:

```make
GOBIN ?= $(GOPATH)/bin
```

and then use $(GOBIN) instead of `$(gopath_bin_path)` below. this has the advantage of respecting any existing `$GOBIN` value in the environment.

PR #125, opened 2019-04-18
[controller] fix pod removal when diff > 1

schallert:
There was a bug in the instance removal logic that would continuously
try to remove the last pod in a set even if it had already been removed
from the placement. This fixes that.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -476,8 +476,11 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 			return fmt.Errorf("set %s has unset spec replica", set.Name)
 		}
 
+		// Number of pods we want in the group.
 		desired := group.NumInstances
+		// Number of pods currently in the group.
 		current := *set.Spec.Replicas
+		// Nunber of instances currently in the placement.
 		inPlacement := int32(len(instancesInIsoGroup(placement, group.Name)))
 
 		setLogger := c.logger.With(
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -476,8 +476,11 @@ func (c *Controller) handleClusterUpdate(cluster *myspec.M3DBCluster) error {
 			return fmt.Errorf("set %s has unset spec replica", set.Name)
 		}
 
+		// Number of pods we want in the group.
 		desired := group.NumInstances
+		// Number of pods currently in the group.
 		current := *set.Spec.Replicas
+		// Nunber of instances currently in the placement.
```
schallert: Fix typo

pkg/controller/update_cluster.go
```diff
@@ -426,55 +430,76 @@ func (c *Controller) expandPlacementForSet(cluster *myspec.M3DBCluster, set *app
...
 	if err != nil {
-		return nil, fmt.Errorf("cannot sort pods: %v", err)
+		return nil, nil, pkgerrors.WithMessage(err, "cannot sort pods")
 	}
 
-	lastPod := podIDs[len(podIDs)-1].pod
+	for i := len(podIDs) - 1; i >= 0; i-- {
+		pod := podIDs[i].pod
+		inst, err := c.findPodInPlacement(cluster, pl, pod)
+		if err == errPodNotInPlacement {
```
schallert: Cause() on error

PR #135, opened 2019-05-02
[health] use dbnode health/bootstrap endpoints

schallert:
Now that M3DB exposes health and bootstrapped endpoints immediately we
can leverage those rather than the crazy health check script.


schallert:
Not to be merged until next release of M3DB.

robskillington:
This should be calling the `/bootstrappedinplacementornoplacement` endpoint I believe?

PR #163, opened 2019-07-17
Add HTTP JSONPB request method to client and update callers

robskillington:
This makes sure that every call site is using the correct JSON marshal/unmarshal and adds helpers so future methods don't have to construct the marshaller themselves and just directly call the `DoHTTPJSONPBRequest` helper.

### DIFF
--- a/pkg/m3admin/client.go
+++ b/pkg/m3admin/client.go
@@ -60,10 +60,6 @@ type client struct {
 	environment string
 }
 
-type nullLogger struct{}
-
-func (nullLogger) Printf(string, ...interface{}) {}
-
 // NewClient returns a new m3admin client.
 func NewClient(clientOpts ...Option) Client {
 	opts := &options{}
@@ -85,7 +81,7 @@ func NewClient(clientOpts ...Option) Client {
 	}
 
 	// We do our own request logging, silence their logger.
-	client.client.Logger = nullLogger{}
+	client.client.Logger.SetOutput(ioutil.Discard)
 	client.client.ErrorHandler = retryhttp.PassthroughErrorHandler
 
 	return client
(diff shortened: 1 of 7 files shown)


pkg/m3admin/client.go
```diff
@@ -167,6 +169,41 @@ func (c *client) DoHTTPRequest(
...
 }
 
+// DoHTTPJSONPBRequest is a helper for performing a request and
+// parsing the response as a JSONPB message into the response.
+// Both request and response are optional and can be emitted if
+// not wanting to either send or receive message.
+func (c *client) DoHTTPJSONPBRequest(
+	action, url string,
+	request proto.Message,
+	response proto.Message,
```
schallert: Just curious, is the reason to pass the response so that a user can pass a typed one rather than have to type assert on the way back (if signature returned `(proto.Message, error)`?

pkg/m3admin/placement/client.go
```diff
@@ -95,41 +87,30 @@ func (p *placementClient) Delete() error {
...
-		return nil, err
-	}
-	if data.Placement == nil {
+
+	if resp.Placement == nil {
 		return nil, errors.New("nil placement fetch")
 	}
-	p.logger.Debug("placement retreived")
-	return m3placement.NewPlacementFromProto(data.Placement)
+	p.logger.Info("placement retreived")
```
schallert: Nit: I changed this from Info to Debug recently because this happens on every run of the control loop. Since we refresh every N seconds even with no events this means the operator logs could just be `"placement retrieved"` constantly.

Might be worth keeping this as debug.

robskillington: Ah yeah, this was a bad merge, will revert.

pkg/m3admin/client.go
robskillington: Well the main thing is I need to unmarshal into a specific proto message, unless the caller wants to do the the whole jsonpb Unmarshal thing (which we're trying to avoid them doing).  That's why they need to provide up front.

Issue #168, opened 2019-08-23
Missing "patch" permission


bai:
* What version of the operator are you running?

`master` at `3d43802a616096f221847026474839f95704139a`.

* What version of Kubernetes are you running? Please include the output of `kubectl version`.

`1.13.7-gke.19`

* What happened?

With https://github.com/m3db/m3db-operator/pull/162 m3db-operator tries to patch pods, which requires additional permissions. Currently `patch` is not on the permissions list.

schallert:
Fixed in #169

PR #169, opened 2019-08-23
Add patch permission to m3db-operator CR

bai:
Add `patch` permission on `pods`, according to https://github.com/m3db/m3db-operator/issues/168.

Might also need to regenerate bundle and remove `update` now that it's replaced by `patch`?


schallert:
Thanks for the fix, @bai!

PR #196, opened 2020-02-14
Apply annotations to pods in created StatefulSets

robskillington:
The StatefulSet template was not applying annotations to pods since it was not part of the pod template. This adds the same annotations to the pod.

### DIFF
--- a/pkg/k8sops/m3db/generators_test.go
+++ b/pkg/k8sops/m3db/generators_test.go
@@ -37,8 +37,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	crdutils "github.com/ant31/crd-validation/pkg"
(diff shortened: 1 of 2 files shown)


pkg/k8sops/m3db/generators_test.go
```diff
@@ -37,8 +37,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	crdutils "github.com/ant31/crd-validation/pkg"
-	"k8s.io/utils/pointer"
 	"github.com/stretchr/testify/assert"
+	"k8s.io/utils/pointer"
```
schallert: Nit since it seems like the import order linter isn't working but can you move this to the other `k8s.io` import groups?

Issue #201, opened 2020-02-24
support hostNetwork for m3db nodes


yuecong:
Thanks for opening an issue for the M3DB Operator! We'd love to help you, but we need the following information included
with any issue:

* What version of the operator are you running? Please include the docker tag. If using `master`, please include the git
  SHA logged when the operator first starts.
v0.5.0

* What version of Kubernetes are you running? Please include the output of `kubectl version`.
v.1.12.10

* What are you trying to do?
We want to enable m3db ndoes able to use [hostNetwork](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces) so that it can boost the pod-to-pod network performance in our k8s environment. Now there is no way to add such customization into m3dbcluster CRD.

* What did you expect to happen?
Able to add [HostNetwotk](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces) and [DnsPolicy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) configuration into the stateful sets created for m3db nodes.

* What happened?
We can not add these two basic configs for m3db nodes.

PR #202, opened 2020-02-24
[api] Add support for hostNetwork

yuecong:
This is to address the proposal as https://github.com/m3db/m3db-operator/issues/201

In this PR, we enabled m3dbCluster CRD able to pass [HostNetwork](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#host-namespaces) and [DnsPolicy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy) to the stateful sets for m3db nodes to boost network performance in some k8s environment. 

I already tested in our environment that the big network improvement for m3db nodes as well as the performance improvement of the whole cluster due to this. Note: All these two configs are not set by default so that for the environment can not benefit from hostNetwork setting, it is a no-op.


schallert:
Thank you for the contribution, @yuecong! I made a few tweaks to tests and updated the docs, but this worked in my test environment. Please let me know if you have any issues with it.

yuecong:
Thanks @schallert !

Issue #204, opened 2020-02-28
operator should only pull the namespace where the operator is in


yuecong:
Thanks for opening an issue for the M3DB Operator! We'd love to help you, but we need the following information included
with any issue:

* What version of the operator are you running? Please include the docker tag. If using `master`, please include the git
0.5.0

* What version of Kubernetes are you running? Please include the output of `kubectl version`.
1.12
* What are you trying to do?
Now the operator is watching all the namespaces for sts and pods. This can be quite slow and also many unnecessary API calls to the API server. e.g. it slows down to patch the pod-identity annotation to each pod. In one of my prod environment, I saw it takes more than 2- 5 minutes to just go through the logic for each pod and added annotations for pod-identity.

* What did you expect to happen?
It should only watch the namespace where the operator is in. I am ok the default behavior to monitor all, but I would like have one option to only monitor specific namespace.

* What happened?
The operator is monitoring all the namespaces and slows down its watch as well as too many unnecessary calls to  k8s api server in a large k8s environment. 

The fix would be very easy, and I already tested in my env. 
It is to change here 
https://github.com/m3db/m3db-operator/blob/master/cmd/m3db-operator/main.go#L171
https://github.com/m3db/m3db-operator/blob/master/cmd/m3db-operator/main.go#L173

to
```
kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient, _informerSyncDuration,kubeinformers.WithNamespace("m3"))

m3dbClusterInformerFactory := informers.NewSharedInformerFactoryWithOptions(crdClient, _informerSyncDuration, informers.WithNamespace("m3"))
```
@schallert if you agree, I would have one PR to add this feature. Basically, I will add one more cmd option so that users can specify whether to monitor all namespaces or just one specific namespace.
By doing this, I can see it is really fast to bootstrap a new m3db cluster.

schallert:
@yuecong do you know if the [`nodeLister`](https://github.com/m3db/m3db-operator/blob/f3b652c29e2e346b68ad60614d3becd5a90de35f/cmd/m3db-operator/main.go#L172) will still work fine even if the informer factory is scoped to a single namespace?

yuecong:
@schallert yes, it does work. I tested in two environments in two different clouds.

[...]

PR #205, opened 2020-02-28
add one option able to only watch one specific namespace

yuecong:
This is to fix the issue proposed as https://github.com/m3db/m3db-operator/issues/204

In this PR,  I add one option so that it can have the operator only watch one namespace rather than all namespaces. In some big k8s environment, it slows the operator to add pod annotation and adds unnecessary high load the API server.

### DIFF
--- a/cmd/m3db-operator/main.go
+++ b/cmd/m3db-operator/main.go
@@ -169,15 +169,22 @@ func main() {
 	}
 
 	stopCh := make(chan struct{})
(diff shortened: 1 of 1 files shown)


xmcqueen:
Merge this thing guys.  It saved me today.  It seems to be working perfectly.  And without it the operator had become nearly non-responsive.

robskillington:
@xmcqueen thank you, looking at this now.

PR #208, opened 2020-03-21
tentative support for an external controlling coordinator

xmcqueen:
This allows the operator to startup in db-only mode (no colocated coordinator).  Uses an externally managed coordinator. Start the db with no "coordinator" section in the configmap.

Pass in the labels of the externally controlled coordinator's service endpoint.

`spec:
.
.
.
  externalCoordinatorSelector:
    app: coordinator
.
.
.`

The db-only cluster will boot up using the external coordinator.


schallert:
Thanks for the contribution, @xmcqueen!

Issue #213, opened 2020-04-30
Documentation should include guidance on when to use `-namespace` flag


SQUIDwarrior:
The documentation should explain how to use the `-namespace` flag, and when to use it.

* What version of the operator are you running? 
0.6.0

* What version of Kubernetes are you running? Please include the output of `kubectl version`.
1.15

* What are you trying to do?
Deploy m3db-operator to a large Kubernetes cluster

* What did you expect to happen?
That it would deploy normally.

* What happened?
With very large clusters, the M3DB nodes would timeout waiting for their pod identity. This is because the m3db-operator is listening for all pod events by default, and it cannot keep up. When this happens, the pod identity annotation is not applied to the M3DB pods before they timeout and restart, causing the cluster to never fully boot. This is fixed by using the `-namespace` flag on the m3db-operator container to filter the events to just the namespace where the M3DB pods are running, but no where in the documentation does it explain that this flag exists, or when to use it.

PR #214, opened 2020-04-30
Added new Production Considerations doc

SQUIDwarrior:
This PR adds a new Production Considerations doc to help address issues outlined here: https://github.com/m3db/m3db-operator/issues/213


xmcqueen:
Looks good! I'm thinking this is not really a "production" config.  Its any time we've got a ton of namespaces.  I'm sure that's the case in large corps with staging areas and test areas.  Even dev areas can get clogged with litter.

Though I do like the "production considerations" concept. Something like "options" seems a more appropriate name.

PR #222, opened 2020-06-09
[controller] Only list objects created by operator

schallert:
Continuously listing every pod and statefulset in large clusters can
slow down the operator control loops significantly. We know every pod
and statefulset created by an M3DBCluster will have predetermined
labels, and can only list objects from the Kubernetes API with those
labels.

### DIFF
--- a/cmd/m3db-operator/main.go
+++ b/cmd/m3db-operator/main.go
@@ -174,7 +174,7 @@ func main() {
 
 	labelReq, err := klabels.NewRequirement(labels.App, selection.Exists, nil)
 	if err != nil {
-		logger.Fatal("failed to construct pod selector requirement", zap.Error(err))
+		logger.Fatal("failed to construct selector requirement", zap.Error(err))
 	}
 	tweakOpts := kubeinformers.WithTweakListOptions(func(opts *metav1.ListOptions) {
 		// PodSelector provides a selector for listing only M3DB pods to annotate.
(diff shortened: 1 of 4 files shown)


cmd/m3db-operator/main.go
```diff
@@ -53,7 +57,7 @@ import (
 
 const (
 	// Informers will resync on this interval
-	_informerSyncDuration = time.Minute
+	_informerSyncDuration = 5 * time.Minute
```
robskillington, line 60: This is a fairly large increase, is it better to take a half measure to 2.5mins or something perhaps?

How often would we see a lost delivery do we think?

cmd/m3db-operator/main.go
```diff
@@ -168,23 +172,42 @@ func main() {
...
+		// filteredInformerFactory is used to create the pod and statefulset
+		// informers. We know that every pod and StatefulSet created by the operator
+		// will have its hardcoded labels. The pod annotation and statefulset loop
+		// may slow down in clusters with many pods or statefulsets if it must
+		// process every objects in the cluster. The filter only lists objects that
+		// have labels we expect.
+		filteredInformerFactory kubeinformers.SharedInformerFactory
+		// The kube informer factory is still required to list nodes, as they don't
+		// have our label.
+		kubeInformerFactory kubeinformers.SharedInformerFactory
```
robskillington, line 196: Do we need a shared informer factory for the node lister?

I checked where `nodeLister` is used and it's only with `.Get(..)` (i.e. lookup a specific host) so would it be possible to just use the kube client for that rather than an informer factory which can be affected by the same issue we're solving here?

cmd/m3db-operator/main.go
schallert: I think it's pretty rare for us to see a lost delivery. We originally had this at a higher value but reduced it to 1min because we thought we were missing events, but it turns out we were overwhelming the controller.

For reference, the default value in kubebuilder is something like 6h and I've never seen issues with code using that.

cmd/m3db-operator/main.go
schallert: If we use the informer then we'll be notified of updates, and our `Get(...)` calls will just read from the cache if it's fresh. If we were to use the client directly it would put slightly more load on the API server.

PR #224, opened 2020-07-12
Comments: add ServiceAccountName to the ClusterSpec

xmcqueen:
add ServiceAccountName to the ClusterSpec for cases where the user does not want to run with the default service account such as pods running under pod security policies

I've not tested this patch on the live clusters yet.  Its likely to work perfectly.  How does it look?


schallert:
Copied these changes over to #225 so I could update generated code before merging. Thanks @xmcqueen!

PR #236, opened 2020-09-25
[controller] Add logic to update StatefulSets

jeromefroe:
This commit to the controller adds support for updating `StatefulSet`'s when their image or config map are out of date with the `ClusterSpec`. After checking that the cluster is healthy, the controller will next check if any `StatefulSet` needs to be updated, and will then perform any needed updates one by one. To ensure that the controller will only perform updates when it is safe to do so, the controller will check for an annotation that must be set by a user to indicate the controller can perform an update.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -32,6 +32,7 @@ import (
 	clientset "github.com/m3db/m3db-operator/pkg/client/clientset/versioned"
 	samplescheme "github.com/m3db/m3db-operator/pkg/client/clientset/versioned/scheme"
 	clusterlisters "github.com/m3db/m3db-operator/pkg/client/listers/m3dboperator/v1alpha1"
+	"github.com/m3db/m3db-operator/pkg/k8sops/annotations"
 	"github.com/m3db/m3db-operator/pkg/k8sops/labels"
 	"github.com/m3db/m3db-operator/pkg/k8sops/m3db"
 	"github.com/m3db/m3db-operator/pkg/k8sops/podidentity"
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -919,3 +948,69 @@ func validateIsolationGroups(cluster *myspec.M3DBCluster) error {
 	}
 	return nil
 }
+
+// shouldUpdateStatefulSet checks if a StatefulSet should be updated. If the Image or
+// ConfigMapName fields in the ClusterSpec don't match what's in the StatefulSet then
+// this function will update the provided StatefulSet to match the desired state and
+// return true.
+func shouldUpdateStatefulSet(spec myspec.ClusterSpec, sts *appsv1.StatefulSet) (bool, error) {
```
schallert, line 956: Image and config updates are definitely the most common case we want to cover, so from that perspective this looks good. There are definitely other cases where the cluster CRD changes and we want to update the statefulsets, and I'm wondering how to address that here. Two thoughts:

1. Any idea what happens if we try to `DeepEqual` the generated spec and the current spec on the cluster? I'm guessing the Kubernetes API makes some sort of transformations that make that not totally possible, but I'm curious.

2. Assuming `DeepEqual` doesn't work, what do you think about a second annotation that will trigger an `Update` regardless of whether image and config has changed?

jeromefroe: I've always had difficulties using `DeepEqual` in the past because the API server sets default (non-zero) values for so many fields. Based on [this `kubebuilder` issue](https://github.com/kubernetes-sigs/kubebuilder/issues/592) though, it seems that we could use [`DeepDerivative`](https://pkg.go.dev/k8s.io/apimachinery/third_party/forked/golang/reflect#Equalities.DeepDerivative) instead since it "is similar to DeepEqual except that unset fields in a1 are ignored ". That issue also mentions [`github.com/banzaicloud/k8s-objectmatcher`](https://github.com/banzaicloud/k8s-objectmatcher) as another alternative if `DeepDerivative` doesn't work for us. If neither package is sufficient though I'm not opposed to having a `force-update` annotation or something similar which will always trigger an update.

pkg/controller/controller.go
```diff
@@ -919,3 +950,57 @@ func validateIsolationGroups(cluster *myspec.M3DBCluster) error {
 	}
 	return nil
 }
+
+// updatedStatefulSet checks if we should update the actual StatefulSet to match it's
+// expected state. The method returns the updated StatefulSet if so, otherwise nil.
+func updatedStatefulSet(
+	actual *appsv1.StatefulSet, cluster *myspec.M3DBCluster, isoGroup myspec.IsolationGroup,
+) (*appsv1.StatefulSet, error) {
```
schallert, line 958: Can we modify the return signature to be `*StatefulSet, bool, error` and use the bool as whether we upgraded rather than implicit `nil`?

schallert (approved):
LGTM 🔥  (w/ 1 comment).

Can do this in another PR if you want but would love to have something in `docs` on using this.

jeromefroe:
> Can do this in another PR if you want but would love to have something in `docs` on using this.

Definitely! I tested the changes on a running cluster yesterday using a local build, but want to deploy the changes and do some more thorough testing today. So I'll update the docs after that once I've gotten a better handle on the UX.

PR #239, opened 2020-09-30
[controller] Create constant for annotation value indicating enabled

jeromefroe:
This commit defines the string that we look for to indicate that an annotation has been enabled (`"enabled"`) as a constant so it can be imported by other packages.


wybczu (approved):
LGTM

schallert (approved):

PR #242, opened 2020-10-07
[api] Allow setting static external coordinator

schallert:
We allow changing the selector that the coordinator service uses,
enabling users to control their M3DB clusters with coordinators that
live outside the cluster. This improves the ability to use
out-of-cluster coordinators, but has the limitation that the coordinator
must be in the same namespace as the cluster.

This PR extends external coordinator support to allow configuring a
static endpoint. This allows the user to manage the cluster with a
coordinator in a different namespace, i.e.
`global-coordinator.$OTHER_NS:7201`.


jeromefroe (approved):
LGTM

PR #245, opened 2020-10-27
Add calls to /namespace/ready if supported by coordinator

nbroyles:
Add support for transitioning namespaces to a ready state after they've been created. This operates as a signal to coordinators pulling namespace information from etcd that the namespace is ready for use.

### DIFF
--- a/pkg/controller/update_cluster.go
+++ b/pkg/controller/update_cluster.go
@@ -73,7 +73,11 @@ func (c *M3DBController) reconcileNamespaces(cluster *myspec.M3DBCluster) error 
 		return err
 	}
 
-	return c.createNamespaces(cluster, resp.Registry)
+	if err := c.createNamespaces(cluster, resp.Registry); err != nil {
+		return err
+	}
+
+	return c.readyNamespaces(cluster, resp.Registry)
 }
 
 // createNamespaces will attempt to create in the cluster all namespaces which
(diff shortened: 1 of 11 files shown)


pkg/controller/update_cluster.go
```diff
@@ -105,6 +110,39 @@ func (c *M3DBController) createNamespaces(cluster *myspec.M3DBCluster, registry
 	return nil
 }
 
+// readyNamespaces will attempt to mark all namespaces in the initializing state as ready.
+func (c *M3DBController) readyNamespaces(cluster *myspec.M3DBCluster, registry *dbns.Registry) error {
+	// NB(nate): Marking namespaces ready (without force) involves checking dbnodes for namespace
+	// availability. External coordinators do not have connections to dbnodes so don't attempt this
+	// step for now.
+	if cluster.Spec.ExternalCoordinator != nil {
```
nbroyles, line 118: FYI, support for this is something we'll be focusing on after 1.0

pkg/controller/update_cluster.go
```diff
@@ -105,6 +110,39 @@ func (c *M3DBController) createNamespaces(cluster *myspec.M3DBCluster, registry
...
+	}
+
+	toReady := namespacesToReady(registry, cluster.Spec.Namespaces)
+	for _, ns := range toReady {
+		req := &admin.NamespaceReadyRequest{
+			Name: ns.Name,
+		}
+		err := c.adminClient.namespaceClientForCluster(cluster).Ready(req)
+		if pkgerrors.Cause(err) == m3admin.ErrNotFound {
+			c.logger.Info("coordinator does not yet support the ready endpoint. " +
```
schallert, line 130: This is probably valuable enough context to keep at `Info`, but just a heads up this will log every time a cluster is processed.

schallert (approved):
LGTM. Let's just make sure we test this in a few different configurations (e.g. external coordinator).

jeromefroe (approved):
LGTM

pkg/controller/update_cluster.go
```diff
@@ -105,6 +109,43 @@ func (c *M3DBController) createNamespaces(cluster *myspec.M3DBCluster, registry
...
+	for _, ns := range toReady {
+		req := &admin.NamespaceReadyRequest{
+			Name: ns.Name,
+		}
+		err := c.adminClient.namespaceClientForCluster(cluster).Ready(req)
+		cause := pkgerrors.Cause(err)
+		// NB(nate): Due to bug in coordinator API routing logic, missing routes may
+		// also be returned as 405s (i.e. method not allowed). So check for that in addition
+		// to 404s.
+		if cause == m3admin.ErrNotFound || cause == m3admin.ErrMethodNotAllowed {
```
nbroyles, line 132: This was a fun little discovery during testing. Depending on the route, misses will either return a `404` or a `405`. Almost certainly a bug in our routing logic, but since these coordinators will be out there in the wild, we need to account for it.

(cc @schallert since this is a small update after the LGTM)

PR #248, opened 2020-11-03
Allow AggregationOptions to be set for a namespace

nbroyles:
AggregationOptions is a new field on NamespaceOptions that allows us to set aggregation details on the namespace definition itself. This is useful for the coordinator when resolving namespaces dynamically instead of via static configuration. This PR allows those same options to be set in the cluster spec.


jeromefroe (approved):
LGTM, @schallert is much more familiar with the codebase than I am though so would be good to get his OK too if there's no rush

schallert (approved):
Nice!

PR #250, opened 2020-11-25
[config] Update default configuration template

wybczu:
This is a breaking change as the operator will *not* be able to create a working default configuration for m3db v1.0.0+. To use the new version of the operator with an older version of m3db you [will have to provide a custom config map](https://operator.m3db.io/configuration/configuring_m3db/).

Please see [m3db changelog](https://github.com/m3db/m3/releases/tag/v1.0.0) for details.


assets/default-config.yaml
```diff
@@ -57,27 +56,26 @@ db:
...
+  filesystem:
     filePathPrefix: /var/lib/m3db
 
-  config:
-    service:
-        env: default_env
+  discovery:
+    config:
+      service:
+        env: "{{ .Env }}"
```
schallert, line 78: This should say as `default_env` since it's not a templated file

schallert (approved):
LGTM pending 1 comment

jeromefroe (approved):

PR #253, opened 2020-12-04
Add support for adding sidecar containers

jeromefroe:
This commit adds support to the operator for adding sidecar containers to the pods that run the M3DB nodes.


schallert (approved):
LGTM. That was easy!

PR #256, opened 2020-12-16
Add configurable port to health client

ryanhall07:
We can't assume the server is listening on port 80


pkg/m3admin/health/client.go
```diff
@@ -31,11 +31,13 @@ import (
 const (
 	bootstrappedPath = "/bootstrapped"
 	m3dbServiceName  = "m3dbnode-m3-cluster-short"
+	defaultPort = 9002
```
schallert, line 34: Nit: reuse consts https://github.com/m3db/m3db-operator/blob/master/pkg/k8sops/m3db/ports.go/#L30

schallert (approved):
LGTM pending const

PR #257, opened 2020-12-16
Use /health endpoint for boostrapped check

ryanhall07:
This returns a 200 instead of an error when not bootstrapped. This makes
it easier to differentiate between transient errors and not
bootstrapped.

### DIFF
--- a/pkg/m3admin/health/client.go
+++ b/pkg/m3admin/health/client.go
@@ -21,7 +21,9 @@
 package health
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 
@@ -30,8 +32,8 @@ import (
 )
 
 const (
-	bootstrappedPath = "/bootstrapped"
-	m3dbServiceName  = "m3dbnode-m3-cluster-short"
+	healthPath      = "/health"
+	m3dbServiceName = "m3dbnode-m3-cluster-short"
 )
 
 type healthClient struct {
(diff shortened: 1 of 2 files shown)


pkg/m3admin/health/client.go
```diff
@@ -30,8 +32,8 @@ import (
 )
 
 const (
-	bootstrappedPath = "/bootstrapped"
-	m3dbServiceName  = "m3dbnode-m3-cluster-short"
+	healthPath      = "/health"
```
nbroyles, line 35: Wouldn't it make more sense to just add this as a check on the client in addition to `/bootstrapped`?

nbroyles: If not, I'd think you'd want to rename client method from `Bootstrapped` to something like `Health`

robskillington (approved):
LGTM

pkg/m3admin/health/client.go
ryanhall07: i don't totally follow. but the problem is this basically would never return false. it would only return true or error.

schallert (approved):

pkg/m3admin/health/client.go
nbroyles: More so suggesting to make it consistent with the API we're calling.  Ideally:
```
HealthClient {
  Bootstrapped() error  // Or just delete this
  Health() (HealthResult, error)
}
```

As it stands, we've got the `Bootstrapped` method invoking the `/health` endpoint which seems inconsistent.

PR #258, opened 2021-01-06
Update README.md

ChrisChinchilla:
I updated the documentation link and then thought that maybe it was easier to direct people to there instead of maintaining two different guides, but I am also open to reproducing the new getting started guide here instead


ChrisChinchilla:
This also means we should update the link in the repository, and possibly delete the _docs_ folder?

jeromefroe (approved):
LGTM, @schallert is much more familiar with the project than I am though so would be good to get his thoughts as well

README.md
```diff
@@ -14,12 +14,6 @@ More information:
...
 
-## Getting Started
-
-The following instructions serve as a quickstart to get an M3DB cluster up and running in your Kubernetes cluster. This
-setup is not for production use, as there's no persistent storage. More information on production-grade clusters can be
-found in our [docs][docs].
-
 ### Kubernetes Cluster Prerequisites
 
 The M3DB operator targets Kubernetes **1.11** and **1.12**. We generally aim to target the latest two minor versions
```
schallert, line 19: Can you also remove this entire paragraph?

schallert (approved):

ChrisChinchilla:
@schallert Removed that paragraph, also removed the docs folder, this might break some build tasks in the main repo, but those tasks should be updated anyway if that's the case ;)

PR #260, opened 2021-01-08
Wait for replica to update pods before processing next statefulset update

robskillington:
This allows for gating progression to updating the next statefulset until all pods have been updated, avoiding the statefulset looking "ready" and "healthy" even though it has a pending update which it's about to process (and start updating pods).


pkg/controller/controller_test.go
```diff
@@ -735,6 +797,30 @@ func TestHandleUpdateClusterUpdatesStatefulSets(t *testing.T) {
...
+				}),
+				newMeta("cluster1-rep2", nil, map[string]string{
+					annotations.Update: "enabled",
+				}),
+			},
+			newImage:                    "m3db:v2.0.0",
+			simulatePodsNotUpdated:      true,
+			expUpdateStatefulSets:       []string{"cluster1-rep0", "cluster1-rep1", "cluster1-rep2"},
+			expFailedUpdateStatefulSets: []string{"cluster1-rep1", "cluster1-rep2"},
+			expNotDone:                  true,
```
wesleyk, line 822: for my own understanding: this would be false before your change?

wesleyk (approved):

pkg/controller/controller.go
```diff
@@ -478,9 +478,34 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
 	// If any of the statefulsets aren't ready, wait until they are as we'll get
 	// another event (ready == bootstrapped)
 	for _, sts := range childrenSets {
-		if sts.Spec.Replicas != nil && *sts.Spec.Replicas != sts.Status.ReadyReplicas {
+		if sts.Spec.Replicas == nil {
 			// TODO(schallert): figure out what to do if replicas is not set
-			c.logger.Info("waiting for statefulset to be ready", zap.String("name", sts.Name), zap.Int32("ready", sts.Status.ReadyReplicas))
+			continue
```
schallert, line 483: Can you just `Warn` here and remove the TODO? I haven't seen this happen in the wild so I don't think we need to figure out how to specifically handle it.

schallert (approved):

pkg/controller/controller_test.go
robskillington: @wesleyk it was more that, before this change if `simulatePodsNotUpdated` had existed in this test code and the "UpdatedReplicas" did not match the expected replicas in the statefulset it would have progressed to update the next statefulset anyway.

pkg/controller/controller.go
robskillington: Yeah sounds good.

jeromefroe (approved):

PR #266, opened 2021-01-15
Add support for pod anti-affinity

### LINKED ISSUE #123: Add flag to enable cluster anti-affinity
Inter-pod [anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature) is a great way to ensure like pods are scheduled on different nodes/failure domains. While it's useful, it also requires "substantial amount of processing which can slow down scheduling in large clusters significantly", so enabling it by default may not be desirable for many users.

We achieve a hacky version of anti-affinity by reserving enough memory for an m3db cluster such that it takes > 50% of a host's memory and no 2 m3db nodes will be scheduled on the same host. Some users may wish to use anti-affinity instead and we should enable that use case for them.

EdgeJ:
Adds functionality to enable pod anti-affinity by matching for the component label against the `componentM3DBNode` value.  This functionality is disabled by default to prevent resource strain on k8s clusters.  This change also splits up the functionality of the `GenerateStatefulSetAffinity` function into two sub-functions handling node affinity and pod anti-affinity.

Fixes #123

### DIFF
--- a/pkg/k8sops/m3db/statefulset_test.go
+++ b/pkg/k8sops/m3db/statefulset_test.go
@@ -188,21 +188,19 @@ func TestGenerateStatefulSetNodeAffinity(t *testing.T) {
 func TestGenerateStatefulSetPodAntiAffinity(t *testing.T) {
 	tests := []struct {
 		isoGroup myspec.IsolationGroup
-		expBool  bool
+		expTerm  string
 		expErr   error
 	}{
 		{
 			isoGroup: myspec.IsolationGroup{
 				Name: "group1",
 			},
-			expBool: false,
 		},
 		{
 			isoGroup: myspec.IsolationGroup{
 				Name:               "group2",
 				UsePodAntiAffinity: false,
 			},
-			expBool: false,
 		},
 		{
 			isoGroup: myspec.IsolationGroup{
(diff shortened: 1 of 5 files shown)


EdgeJ:
Converted to draft, as I realized I forgot to handle `topologyKey` to match node labels.

pkg/k8sops/m3db/statefulset_test.go
```diff
@@ -181,3 +184,69 @@ func TestGenerateStatefulSetAffinity(t *testing.T) {
...
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      labels.Component,
+					Operator: "In",
+					Values:   []string{labels.ComponentM3DBNode},
+				},
+			},
+		}
+
+		assert.Equal(t, expTerms, terms[0].LabelSelector)
```
schallert, line 250: Can you also verify that `PodAffinityToplogyKey` is set to the expected value in the tests?

pkg/k8sops/m3db/statefulset_test.go
```diff
@@ -181,3 +184,69 @@ func TestGenerateStatefulSetAffinity(t *testing.T) {
...
+
+		if !test.expBool {
+			assert.Nil(t, antiaffinity)
+			continue
+		}
+
+		if test.expErr != nil {
+			assert.Equal(t, test.expErr, err)
+			continue
+		}
```
schallert, line 236: Nit: can you check this before `expBool` so that we don't miss an error just because of the bool output?

EdgeJ: Updated the tests to remove the use of `expBool` and simply check the value of `test.isoGroup.UsePodAntiAffinity`, since that's the input we want to be testing anyway.

schallert (approved):

schallert:
Thanks for the contribution!

PR #275, opened 2021-02-25
[controller] Support multi instance placement add

notbdu:
Modifies the placement client `Add()` API to accept multiple instances.

### DIFF
--- a/pkg/controller/update_cluster.go
+++ b/pkg/controller/update_cluster.go
@@ -21,7 +21,6 @@
 package controller
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -46,7 +45,6 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	klabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
 
@@ -103,7 +101,7 @@ func (c *M3DBController) createNamespaces(cluster *myspec.M3DBCluster, registry 
 				zap.String("namespace", ns.Name),
 				zap.Error(err))
 
-			return fmt.Errorf("error creating namespace '%s': %v", ns.Name, err)
+			return fmt.Errorf("error creating namespace '%s': %w", ns.Name, err)
 		}
 
 		c.recorder.NormalEvent(cluster, eventer.ReasonCreating, "created namespace "+ns.Name)
@@ -298,8 +296,8 @@ func (c *M3DBController) setStatusPlacementCreated(cluster *myspec.M3DBCluster) 
 	var err error
 	cluster, err = c.crdClient.OperatorV1alpha1().M3DBClusters(cluster.Namespace).UpdateStatus(cluster)
 	if err != nil {
-		err := fmt.Errorf("error updating cluster placement init status: %v", err)
-		c.logger.Error(err.Error())
+		c.logger.Error("error updating cluster placement init status",
+			zap.Error(err))
 		c.recorder.WarningEvent(cluster, eventer.ReasonFailSync, "failed to update placement status: %v", err)
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/controller/update_cluster.go
```diff
@@ -355,43 +356,52 @@ func (c *M3DBController) reconcileBootstrappingStatus(cluster *myspec.M3DBCluste
...
+		instances    = make([]*placementpb.Instance, 0, len(pods))
+		reasonBuf    = bytes.NewBufferString("adding pods to placement (")
+		loggerFields = make([]zap.Field, 0, len(pods))
+	)
+	for _, pod := range pods {
+		c.logger.Info("found pod not in placement", zap.String("pod", pod.Name))
+		inst, err := m3db.PlacementInstanceFromPod(cluster, pod, c.podIDProvider)
+		if err != nil {
+			err := fmt.Errorf("error creating instance for pod %s", pod.Name)
+			c.logger.Error(err.Error())
```
schallert, line 374: Nit: take advantage of structured logging here (i.e. `logger.Error("error creating instance for pod", zap.String("pod", ...`

pkg/controller/update_cluster.go
```diff
@@ -355,43 +356,52 @@ func (c *M3DBController) reconcileBootstrappingStatus(cluster *myspec.M3DBCluste
...
-	c.logger.Info("found pod not in placement", zap.String("pod", pod.Name))
-	inst, err := m3db.PlacementInstanceFromPod(cluster, pod, c.podIDProvider)
-	if err != nil {
-		err := fmt.Errorf("error creating instance for pod %s", pod.Name)
-		c.logger.Error(err.Error())
-		return err
+func (c *M3DBController) addPodsToPlacement(cluster *myspec.M3DBCluster, pods []*corev1.Pod) error {
+	var (
+		instances    = make([]*placementpb.Instance, 0, len(pods))
+		reasonBuf    = bytes.NewBufferString("adding pods to placement (")
```
schallert, line 366: This is a pretty different convention than how build up logs, and this isn't very perf-sensitive code.

I would vote for just adding to an array of strings, which are the names of the instances / pods, and then calling `zap.Strings`.

pkg/controller/update_cluster.go
```diff
@@ -355,43 +356,52 @@ func (c *M3DBController) reconcileBootstrappingStatus(cluster *myspec.M3DBCluste
...
+		err := fmt.Errorf("error setting pods bootstrapping status: %w", err)
 		c.logger.Error(err.Error())
 		return err
 	}
 
-	err = c.adminClient.placementClientForCluster(cluster).Add(*inst)
+	err = c.adminClient.placementClientForCluster(cluster).Add(instances)
 	if err != nil {
-		err := fmt.Errorf("error adding pod to placement: %s", pod.Name)
+		err := fmt.Errorf("error: %s", reason)
```
schallert, line 392: Same thing here re: taking the change to structure the below `.Error` log.

pkg/controller/update_cluster.go
```diff
@@ -38,12 +39,14 @@ import (
...
 	"github.com/m3db/m3db-operator/pkg/util/eventer"
 
+	"github.com/m3db/m3/src/cluster/generated/proto/placementpb"
 	"github.com/m3db/m3/src/cluster/placement"
 	dbns "github.com/m3db/m3/src/dbnode/generated/proto/namespace"
 	"github.com/m3db/m3/src/query/generated/proto/admin"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
schallert, line 49: Nit: this ls already imported as `corev1` on the line above.

schallert (approved):
LGTM. A few minor nits.

One general question: do our current "ready replica" checks guarantee that we'll add all pods at once, rather than accidentally adding one at a time?

For example, say we scale a set from 2 to 4 pods. At time T1, 1/2 new pods is up. Are our readiness checks structured in such a way that we won't immediate add the 3rd pod, and will wait until we can evaluate the two new ones at once?

notbdu:
In practice, I observed the operator bringing up a pod at a time (waiting for each to be ready before proceeding) and then expanding the placement when all requested pods are up/ready/bootstrapped (w/ no shards).

In the code it looks like we add a pod at a time here:
https://github.com/m3db/m3db-operator/blob/master/pkg/controller/controller.go#L710-L716

And then when the number of `desired == current` pods we expand the placement here:
https://github.com/m3db/m3db-operator/blob/master/pkg/controller/controller.go#L688-L701

PR #276, opened 2021-02-25
[ci] Update kind; fix node image

schallert:
Also make test manifests work better out of the box with kind.

### DIFF
--- a/example/etcd/etcd-kind.yaml
+++ b/example/etcd/etcd-kind.yaml
@@ -1,0 +1,103 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: etcd
+  labels:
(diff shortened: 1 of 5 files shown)


wesleyk (approved):

example/etcd/etcd-kind.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: v1
```
wesleyk, line 1: is this etcd kind example sourced from somewhere? Could we link the source if so?

schallert: It's our own thing. We have some static manifests for etcd in this directory, and I just took those and added a storageclass.

Issue #278, opened 2021-03-01
Migrate from gometalinter to golangci-lint


wybczu:
See https://github.com/m3db/m3/issues/2558 for details.

PR #279, opened 2021-03-01
[ci] Switch to golangci-lint

wybczu:
As we're no longer use metalinter in m3 monorepo it makes sense to switch to golangci-lint here too.

Fixes https://github.com/m3db/m3db-operator/issues/278.


.golangci.yml
```diff
@@ -0,0 +1,283 @@
...
+  exclude-use-default: false
+
+  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  max-issues-per-linter: 0
+
+  # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
+  max-same-issues: 0
+
+  # Show only new issues created after git revision `REV`
+  new-from-rev: a0ecbf9d10486673d7581f4e6a8b0f30497f8a66
```
schallert, line 283: How many errors are there if we remove this? I.e. is it worth it to bite the bullet now and fix them?

schallert: Never mind, it looks pretty gnarly. This SGTM.

schallert (approved):

.golangci.yml
wybczu: Around 600 errors. Top 5 linters by error count:

```
 138 paralleltest
 115 gci
 105 lll
  44 errcheck
  30 errorlint
```

PR #281, opened 2021-03-02
Make sure rollout restarts are performed for stateful sets

soundvibe:
It would be really useful if operator would be able to initiate rolling restart for stateful set when update annotation is set even when there are no changes in cluster spec. Since this would be done only when update annotation is set, there shouldn't be any unexpected outcomes.

Currently, rollout updates won't be initiated because in order for them to be performed, spec pod's template needs to change. 
This PR ensures that spec pod's template is always updated when update annotation is set.


jeromefroe (commented):
Changes look good to me overall! The only question I have is if we want to make the rollouts configurable via a cluster's spec? Specifically, would it be worth requiring that the user enables rollouts via a field in the spec so that the current behavior remains the default?

pkg/controller/controller.go
```diff
@@ -1083,9 +1086,18 @@ func updatedStatefulSet(
...
+	updateSpecTemplate(expected)
 	return expected, true, nil
 }
 
+// updates spec template so that rollout update is always performed.
+func updateSpecTemplate(sts *appsv1.StatefulSet) {
+	if sts.Spec.Template.Annotations == nil {
+		sts.Spec.Template.Annotations = map[string]string{}
+	}
+	sts.Spec.Template.Annotations[annotations.Rollout] = time.Now().Format(rolloutTimeFormat)
```
schallert, line 1098: Nit: can you use the k8s-native [time helpers](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Now)? That should handle formatting for you.

soundvibe: Will do. I assume k8s annotation values don't have similar restrictions as label values?

soundvibe:
What's the status on this? Could it be merged? Maybe rollouts through cluster's spec could be implemented as a separate PR later?

pkg/controller/controller.go
```diff
@@ -1304,9 +1305,20 @@ func updatedStatefulSet(
...
 }
 
+// updates spec template so that rollout update is always performed.
+func updateSpecTemplate(sts *appsv1.StatefulSet) {
+	if sts.Spec.Template.Annotations == nil {
+		sts.Spec.Template.Annotations = map[string]string{}
+	}
+
+	now, _ := metav1.Now().MarshalQueryParameter()
+	sts.Spec.Template.Annotations[annotations.Rollout] = now
```
jeromefroe, line 1319: I wonder if we should only do this if the user opts in to this behavior by setting a corresponding field in the cluster spec? I just worry it might come as a surprise to someone that was expecting the old behavior of the operator not performing an update if the statefulsets hadn't changed.

soundvibe: You still need to update sts with `update` annotation for rollout to happen, I am assuming that if someone already added this annotation they probably expect that sts should be updated. Otherwise what's the point for adding an `update` annotation in the first place?

jeromefroe: The main time I can think of that having the operator skip statefulsets that don't need to be updated has come in handy is when an update to a cluster has been interrupted partway through. For example, we would set the annotation on all statefulsets, but then if we ran into an issue we could remove the annotation from all statefulsets to ensure the operator doesn't perform any more updates while we debug any issue that we encountered. Once we've decided to continue with the update we could once again set the annotation on all statefulsets and leave it up to the operator to determine which statefulsets still need to be updated and which don't, skipping any nodes which have already been updated. It's admittedly not a very common use case but I worry that if anyone has become used to it then they may be surprised if the behavior changes here. But I'm also not convinced that we necessarily need to keep supporting this behavior. @schallert what do you think?

PR #282, opened 2021-03-03
Support namespace ExtendedOptions in cluster spec

linasm:
This adds support for namespace `ExtendedOptions` field in cluster spec map (passing its value to namespace create request, if present).

Note that I have used intermediary transformation via json (from cluster spec to protobuf) to avoid implementing manual, error prone conversion of dynamic `map[string]interface{}` which is used in `ExtendedOptions`.


pkg/apis/m3dboperator/v1alpha1/namespace.go
```diff
@@ -98,6 +100,37 @@ type DownsampleOptions struct {
...
+}
+
+func (o *dynamicOption) MarshalJSON() ([]byte, error) {
+	return json.Marshal(o.o)
+}
+
+// ExtendedOptions stores the extended namespace options.
+type ExtendedOptions struct {
+	Type    string                   `json:"type,omitempty"`
+	Options map[string]DynamicOption `json:"options,omitempty"`
```
schallert, line 131: I've never seen interface types in Kubernetes API types before. I wonder if having this be `map[string]json.RawMessage` would give you the same benefits (auto-generated DeepCopy implementation, more control over serialization) while being more Kubernetes-friendly?

linasm: Thanks, @schallert , `json.RawMessage` really makes the implementation much simpler. Updated.

schallert (approved):
LGTM, but it would be good to test on a live operated cluster before we merge, both with and without extended options. I've seen some... "unintuitive" (to put it nicely) behavior from Kubernetes before when serializing / deserializing anything non-standard.

linasm:
> LGTM, but it would be good to test on a live operated cluster before we merge, both with and without extended options. I've seen some... "unintuitive" (to put it nicely) behavior from Kubernetes before when serializing / deserializing anything non-standard.
Sure, we've done some testing on a real cluster.

PR #283, opened 2021-03-11
Add support for parallel node updates within a statefulset

nbroyles:
This commits allows multiple nodes to be updated and deployed
at the same time within a statefulset. Opting into this behavior
requires the user to update the ClusterSpec to set OnDeleteUpdateStrategy
to true. When this is set, the operator will watch for the
operator.m3db.io/parallel-update annotation on statefulsets. When found,
the operator will attempt to update each statefulset, rolling N many
pods at a time where N is the value of the parallel-update annotation.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -790,7 +790,7 @@ func (c *M3DBController) patchStatefulSet(
 		StatefulSets(set.Namespace).
 		Patch(set.Name, types.MergePatchType, patchBytes)
 	if err != nil {
-		return fmt.Errorf("error updating statefulset %s: %v", set.Name, err)
+		return fmt.Errorf("error updating statefulset %s: %w", set.Name, err)
 	}
 
 	return nil
@@ -863,22 +863,22 @@ func (c *M3DBController) updateStatefulSetNodes(
 		}
 		logger.Info("restarting pods", zap.Any("pods", names))
 		return true, nil
-	} else {
-		// If there are no pods to update, we're fully rolled out so remove
-		// the update annotation.
-		if err = c.patchStatefulSet(sts, func(set *appsv1.StatefulSet) {
-			delete(set.Annotations, annotations.ParallelUpdateInProgress)
-
-			// NB(nate): K8s handles this for you when using the RollingUpdate update strategy.
-			// However, since OnDelete removes k8s from the pod update process, it's our
-			// responsibility to set this once done rolling out.
-			set.Status.CurrentReplicas = set.Status.UpdatedReplicas
-			set.Status.CurrentRevision = set.Status.UpdateRevision
-		}); err != nil {
-			return false, err
-		}
-		logger.Info("update complete")
 	}
+
+	// If there are no pods to update, we're fully rolled out so remove
+	// the update annotation.
+	if err = c.patchStatefulSet(sts, func(set *appsv1.StatefulSet) {
+		delete(set.Annotations, annotations.ParallelUpdateInProgress)
+
+		// NB(nate): K8s handles this for you when using the RollingUpdate update strategy.
+		// However, since OnDelete removes k8s from the pod update process, it's our
+		// responsibility to set this once done rolling out.
+		set.Status.CurrentReplicas = set.Status.UpdatedReplicas
+		set.Status.CurrentRevision = set.Status.UpdateRevision
+	}); err != nil {
+		return false, err
+	}
+	logger.Info("update complete")
 
 	return false, nil
 }
@@ -933,7 +933,7 @@ func (c *M3DBController) podsToUpdate(
 		return nil, err
 	}
 
-	var toUpdate []*corev1.Pod
+	toUpdate := make([]*corev1.Pod, 0, len(sortedPods))
 	for _, pod := range sortedPods {
 		toUpdate = append(toUpdate, pod.pod)
 		if len(toUpdate) == numPods {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -920,6 +920,7 @@ func (c *M3DBController) podsToUpdate(
 		currRev   = sts.Status.CurrentRevision
 		updateRev = sts.Status.UpdateRevision
 	)
+	// nolint:gocritic
 	if currRev == "" {
 		return nil, errors.New("currentRevision empty")
 	} else if updateRev == "" {
@@ -930,19 +931,10 @@ func (c *M3DBController) podsToUpdate(
 	}
 
 	// Get any pods not on the updateRevision for this statefulset
-	revReq, err := klabels.NewRequirement(
-		"controller-revision-hash", selection.NotEquals, []string{updateRev},
-	)
+	podSelector, err := generatePodSelector(updateRev, sts)
 	if err != nil {
 		return nil, err
 	}
-	stsReq, err := klabels.NewRequirement(
-		labels.StatefulSet, selection.Equals, []string{sts.Name},
-	)
-	if err != nil {
-		return nil, err
-	}
-	podSelector := klabels.NewSelector().Add(*revReq, *stsReq)
 	pods, err := c.podLister.Pods(namespace).List(podSelector)
 	if err != nil {
 		return nil, err
@@ -966,6 +958,23 @@ func (c *M3DBController) podsToUpdate(
 	}
 
 	return toUpdate, nil
+}
+
+func generatePodSelector(updateRev string, sts *appsv1.StatefulSet) (klabels.Selector, error) {
+	revReq, err := klabels.NewRequirement(
+		"controller-revision-hash", selection.NotEquals, []string{updateRev},
+	)
+	if err != nil {
+		return nil, err
+	}
+	stsReq, err := klabels.NewRequirement(
+		labels.StatefulSet, selection.Equals, []string{sts.Name},
+	)
+	if err != nil {
+		return nil, err
+	}
+	podSelector := klabels.NewSelector().Add(*revReq, *stsReq)
+	return podSelector, nil
 }
 
 func instancesInIsoGroup(pl m3placement.Placement, isoGroup string) []m3placement.Instance {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 	"sync"
 
 	"github.com/m3db/m3db-operator/pkg/apis/m3dboperator"
@@ -57,12 +58,12 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/pointer"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	pkgerrors "github.com/pkg/errors"
 	"github.com/uber-go/tally"
 	"go.uber.org/zap"
-	"k8s.io/utils/pointer"
 )
 
 const (
@@ -315,7 +316,6 @@ func (c *M3DBController) processClusterQueueItem() bool {
 
 		return nil
 	}(obj)
-
 	if err != nil {
 		runtime.HandleError(err)
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/controller_test.go
```diff
@@ -228,14 +249,15 @@ func waitForStatefulSets(
...
 		}
+		if len(updatedPods) > 0 {
+			podUpdateGroups = append(podUpdateGroups, updatedPods)
+			updatedPods = nil
+		}
 		mu.Unlock()
 
 		if seen != len(opts.expectedStatefulSets) {
 			time.Sleep(100 * time.Millisecond)
-			continue
```
nbroyles, line 235: This continue and break logic actually caused the test method to exit before statefulset updates were finished so just remove and rely on the iterations limit.

pkg/controller/controller.go
```diff
@@ -765,6 +741,220 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+
+	numPods, err := c.getMaxPodsToUpdate(actual)
+	if err != nil {
+		logger.Error("error retrieving number of pods to update in parallel", zap.Error(err))
+		return false, err
+	}
+
+	if numPods == 0 {
+		err = errors.New("parallel update annotation set to 0. will not perform pod updates")
+		logger.Error(err.Error())
```
ryanhall07, line 853: is it really necessary to log the errors in this method? I would think something higher up would be logging the errors

pkg/controller/controller.go
```diff
@@ -765,6 +741,220 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+		zap.String("expected_currentRevision", expected.Status.CurrentRevision),
+		zap.String("expected_updateRevision", expected.Status.UpdateRevision),
+		zap.Int64("generation", expected.Generation),
+		zap.Int64("observed", expected.Status.ObservedGeneration),
+	)
+
+	return false, err
+}
+
+func (c *M3DBController) updateWithOnDeleteStrategy(
```
ryanhall07, line 809: it would be nice to document the semantics of the returned boolean

pkg/controller/controller.go
```diff
@@ -765,6 +741,220 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+			zap.String("actual_currentRevision", actual.Status.CurrentRevision),
+			zap.String("actual_updateRevision", actual.Status.UpdateRevision),
+			zap.Int32("expected_readyReplicas", expected.Status.ReadyReplicas),
+			zap.Int32("expected_updatedReplicas", expected.Status.UpdatedReplicas),
+			zap.String("expected_currentRevision", expected.Status.CurrentRevision),
+			zap.String("expected_updateRevision", expected.Status.UpdateRevision),
+			zap.Int64("generation", expected.Generation),
+			zap.Int64("observed", expected.Status.ObservedGeneration),
+		)
+	} else if _, ok := actual.Annotations[annotations.ParallelUpdateInProgress]; !ok {
```
ryanhall07, line 840: it seems like this method is doing a lot with several different flows. is there anyway we can break this up? 

it seems updating the sts is the same between the 2 different rollout strategies, so that's probably something we could break out.

ryanhall07 (commented):
generally LG

pkg/controller/controller.go
```diff
@@ -560,29 +560,23 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
 			return err
 		}
 
-		if !update {
-			continue
+		var processNext bool
+		// The expectation is that these update methods will return false when they have updated
+		// actual pods in the statefulset. This will stop any further processing of pods until
+		// the updated ones are back healthy
+		if cluster.Spec.OnDeleteUpdateStrategy {
```
jeromefroe, line 567: Do we know if we were to update a `StatefulSet`'s `template` and its `updateStrategy` if the controller will use the new `updateStrategy` to update the set or the old one? Just wondering about a potential edge case where we update both, and the StatefulSet controller used the old strategy for the update but because we're checking `cluster.Spec.OnDeleteUpdateStrategy` here we expect the new one will be used. I would _think_ the controller would use the new strategy that's in the update but Kube has surprised me before 😄

pkg/controller/controller.go
nbroyles: Yeah, this is generally the pattern I observed in the operator. Traced the call stack to where these errors are handled and nothing does.

ryanhall07: yea nvm, this is cause we don't propagate call stacks in the operator.

jeromefroe (approved):
LGTM! Just had one edge question on how the StatefulSet controller handles changes to a set's update strategy.

pkg/controller/controller.go
nbroyles: Actually, nevermind. Will move this logging up to the loop that invokes the updates depending on the update strategy which should help a bit

pkg/controller/controller.go
nbroyles: Yeah, simplified this to make it a little more clear.

pkg/controller/controller.go
```diff
@@ -765,6 +741,220 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+	return nil
+}
+
+func (c *M3DBController) updateWithRollingUpdateStrategy(
+	cluster *myspec.M3DBCluster,
+	actual *appsv1.StatefulSet,
+	expected *appsv1.StatefulSet,
+	update bool,
+) (bool, error) {
+	if !update {
```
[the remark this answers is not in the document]
nbroyles: We actually still want to call `updateWithOnDeleteStrategy` even if we're not updating the statefulset as we may still need to roll nodes in the set. Simplified the logic to make this a little more clear

pkg/controller/controller.go
nbroyles: Yeah, so I just tested this out and it does respect the new `updateStrategy`. So when you change the template and set `OnDelete`, k8s (correctly, imo) waits for you to roll the statefulset pods to pick up the update.

pkg/controller/controller.go
```diff
@@ -560,29 +560,46 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+		// updates as we want to wait for those updates to be completed successfully first
+		// (i.e. new pods are in the ready state).
+		processNext := true
+		if update {
+			actual, err = c.applyStatefulSetUpdate(cluster, actual, expected)
+			if err != nil {
+				c.logger.Error(err.Error())
+				return err
+			}
+			processNext = false
```
schallert, line 583: Rather than have `processNext = false` here and continue this loop, can we just `return nil` so that we exit + re-trigger the entire reconciliation loop? That way we'll run all of our safety checks again.

schallert: AFAICT this would also match the old post-update behavior.

pkg/controller/controller.go
```diff
@@ -765,6 +764,186 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+		names := make([]string, 0, len(pods))
+		for _, pod := range pods {
+			if err := c.kubeClient.CoreV1().
+				Pods(pod.Namespace).
+				Delete(pod.Name, &metav1.DeleteOptions{}); err != nil {
+				return false, err
+			}
+			names = append(names, pod.Name)
+		}
+		logger.Info("restarting pods", zap.Any("pods", names))
```
schallert, line 864: Nit: you can use `zap.Strings` here since `names` is a `[]string`.

pkg/controller/controller.go
```diff
@@ -765,6 +764,186 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+	// If there are no pods to update, we're fully rolled out so remove
+	// the update annotation.
+	if err = c.patchStatefulSet(sts, func(set *appsv1.StatefulSet) {
+		delete(set.Annotations, annotations.ParallelUpdateInProgress)
+
+		// NB(nate): K8s handles this for you when using the RollingUpdate update strategy.
+		// However, since OnDelete removes k8s from the pod update process, it's our
+		// responsibility to set this once done rolling out.
+		set.Status.CurrentReplicas = set.Status.UpdatedReplicas
+		set.Status.CurrentRevision = set.Status.UpdateRevision
```
schallert, line 877: This is surprising / also concerning to have to reconcile k8s' status for it. Are there docs somewhere on this?

If we have to do this, I would also prefer we use `Update` here rather than a patch. That way if k8s has modified the statefulset in between when we decided to change this field + when we made the patch, we'll get a conflict and have to re-reconcile.

pkg/controller/controller.go
```diff
@@ -765,6 +764,186 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+	}); err != nil {
+		return false, err
+	}
+	logger.Info("update complete")
+
+	return false, nil
+}
+
+func (c *M3DBController) getMaxPodsToUpdate(actual *appsv1.StatefulSet) (int, error) {
```
schallert, line 886: Can this be a free function rather than a method on `*M3DBController`?

pkg/controller/controller.go
nbroyles: Yeah, it's very unfortunate. I pieced it together from a bug report (https://github.com/kubernetes/kubernetes/issues/73492) and looking at the update code (https://github.com/kubernetes/kubernetes/blob/053e7f4b0430e1f3814bb16bdb506cd59bb26593/pkg/controller/statefulset/stateful_set_utils.go#L377). But definitely let me know if I'm off base here.

pkg/controller/controller.go
nbroyles: lol, yes, that's much simpler. Will update.

pkg/controller/controller.go
nbroyles: Nice, TIL!

pkg/controller/controller.go
nbroyles: Actually, wait. Only, if the updateStrategy is rolling update can we do that. Still probably nice to bail early if using rolling update, so will modify for that.

nbroyles: OK, pushed a change to clarify this logic even more. It's worth noting that none of the semantics around this loop have changed here. In both the RollingUpdate and OnDelete cases, if changes are made to a statefulset, the operator exits this loop and relies on event updates (i.e. pod ready) to continue updating.

pkg/controller/controller.go
nbroyles: Also, was looking closer at the StatefulSetInterface and saw this method:

```go
	UpdateStatus(*v1.StatefulSet) (*v1.StatefulSet, error)
```

What are your thoughts on this change then?

```go
	sts.Status.CurrentReplicas = sts.Status.UpdatedReplicas
	sts.Status.CurrentRevision = sts.Status.UpdateRevision

	if sts, err = c.kubeClient.AppsV1().StatefulSets(cluster.Namespace).UpdateStatus(sts); err != nil {
		return false, err
	}

	if err = c.patchStatefulSet(sts, func(set *appsv1.StatefulSet) {
		delete(set.Annotations, annotations.ParallelUpdateInProgress)
	}); err != nil {
		return false, err
	}
```

Would that also generate a conflict since it's no longer patching for the revision update? Also, seems safe to patch to remove the annotation, but correct me if I'm wrong.

schallert: This SGTM 👍  .

schallert (approved):
LGTM. Very nice! The changes make this easy to follow.

pkg/controller/controller.go
```diff
@@ -765,6 +766,191 @@ func (c *M3DBController) handleClusterUpdate(cluster *myspec.M3DBCluster) error
...
+		zap.String("expected_currentRevision", expected.Status.CurrentRevision),
+		zap.String("expected_updateRevision", expected.Status.UpdateRevision),
+		zap.Int64("generation", expected.Generation),
+		zap.Int64("observed", expected.Status.ObservedGeneration),
+	)
+	return updated, nil
+}
+
+// updateStatefulSetNodes returns true if it updates any pods
+func (c *M3DBController) updateStatefulSetNodes(
```
ryanhall07, line 829: nit: it would be nice to be consistent with nodes vs pods in the terminology.

ryanhall07: i lean towards pods since this is k8s code

ryanhall07 (approved):
thanks for iterating. LG!

PR #286, opened 2021-04-26
[k8sops] Remove liveness probes from DB

schallert:
We already use readiness probes to make sure dbnodes are bootstrapped
before taking various actions. Liveness probes are usually a good idea,
but in the case of the DB we'd rather a deadlocked node remain up so we
can debug vs. automatically killing it.


robskillington (approved):
LGTM

ryanhall07 (approved):

linasm (approved):

PR #304, opened 2021-10-20
[deps] Update k8s.io/kube-openapi dependency

jeromefroe:
This commit updates our dependency on `k8s.io/kube-openapi`. In a recent commit, the package made a breaking change to the `OpenAPIDefinition` type but since the package doesn't use tagged releases, the Go toolchain has trouble including the package when a module has a transitive dependency on different versions of the package. By updating to the latest commit we can ensure that as long as packages of the `m3db-operator` can update their other dependencies to the latest version as well then the latest version will work for the package.


schallert (approved):
LGTM

PR #314, opened 2022-03-07
batch node removal

Antanukas:
Adds support to shrink cluster by multiple instances


pkg/controller/update_cluster.go
```diff
@@ -556,25 +556,36 @@ func (c *M3DBController) shrinkPlacementForSet(
...
+	for idx, inst := range removeInst {
+		removeIds[idx] = inst.ID()
+	}
+	c.logger.Info("removing instances from placement",
+		zap.String("instances", strings.Join(removeIds, ",")))
+	return c.adminClient.placementClientForCluster(cluster).Remove(removeIds)
 }
 
-// findPodInstanceToRemove returns the pod (and associated placement instace)
+// findPodInstancesToRemove returns the pod (and associated placement instace)
```
jeromefroe, line 579: super nit: `s/pod/pods`

jeromefroe (approved):

soundvibe (approved):

pkg/controller/update_cluster.go
schallert, line 579: Nit: can you update the comment to reflect that this will return multiple pods / instances now?

pkg/controller/update_cluster.go
```diff
@@ -595,10 +611,11 @@ func (c *M3DBController) findPodInstanceToRemove(
...
 			return nil, nil, pkgerrors.WithMessage(err, "error finding pod in placement")
 		}
-		return pod, inst, nil
+		leftToRemove--
+		podsToRemove = append(podsToRemove, pod)
+		instancesToRemove = append(instancesToRemove, inst)
 	}
-
-	return nil, nil, errNoPodsInPlacement
+	return podsToRemove, instancesToRemove, nil
```
schallert, line 618: To ensure this is doing what the user intends, do we want to return an error if `len(podsToRemove) != removeCount`? Not sure if there's a case where that could happen.

schallert (approved):
LGTM

pkg/controller/update_cluster.go
Antanukas: not sure, podsToRemove seems to only be used in tests for assertions. We ignore it in production code.

Antanukas: Also I don't think that can happen with current impl, since both arrays are only appended under same conditions.

Antanukas: I think the only case where this can happen is if we somehow want to remove more than exists. There is a test case when removeCount is for 4 but we have only 3 instances in placement.

In practice this should not happen because we remove only if placement contains more than desired:
```
		if inPlacement > desired {
			setLogger.Info("remove instance from placement for set")
			return c.shrinkPlacementForSet(cluster, set, placement, int(inPlacement-desired))
		}
```

. Maybe if we spam expand/shrink we could enter some weird state here. But I think it's hard to reason what will happen anyways. 

Another approach that potentially can be safer is to instead of having "removeCount" we could pass in "desired" so that we try to reach target state, we could than assert if target state makes sense like "desired" > "currentPodCount" otherwise it looks like expansion.

Antanukas:
Refactored PR to instead of using `removeCount == inPlacement - desired` just pass `desired` to shrink method that tries to reach the desired count. If desired is negative it just bails. If current state is lower or equal to desired it will not do any shrinking. I find this approach easier to understand and reason about.

pkg/controller/update_cluster.go
```diff
@@ -584,21 +599,25 @@ func (c *M3DBController) findPodInstanceToRemove(
 		return nil, nil, pkgerrors.WithMessage(err, "cannot sort pods")
 	}
 
-	for i := len(podIDs) - 1; i >= 0; i-- {
+	var (
+		podsToRemove      []*corev1.Pod
+		instancesToRemove []placement.Instance
+	)
+	for i := len(podIDs) - 1; i >= desiredInstanceCount; i-- {
```
jeromefroe, line 606: nit: Maybe it's worth moving the check that `desiredInstanceCount` is greater than zero into this function just in case we call it from other places in the future?

jeromefroe (approved):
LGTMx2, I like the new approach of just passing the desired number of instances to `shrinkPlacementForSet` better as well

pkg/controller/update_cluster.go
```diff
@@ -556,49 +557,68 @@ func (c *M3DBController) shrinkPlacementForSet(
...
 	pl placement.Placement,
 	pods []*corev1.Pod,
-) (*corev1.Pod, placement.Instance, error) {
+	desiredInstanceCount int,
+) ([]*corev1.Pod, []placement.Instance, error) {
 	if len(pods) == 0 {
 		return nil, nil, errEmptyPodList
 	}
+	if desiredInstanceCount < 0 {
+		msg := fmt.Sprintf("desired instance count is negative: %d", desiredInstanceCount)
```
linasm, line 594: Shouldn't we also validate against `desiredInstanceCount = 0`?

Antanukas: I think 0 should be a valid optin. Like we can do with `kubectl set scale 0`

linasm: How can it be valid in terms of placement - where will the shards be relocated to?

Antanukas: Hmm yeah maybe it does not make sense after all.

PR #316, opened 2022-03-15
prevent scaling down to 0 instances

Antanukas:
After discussing with Linas we agreed that scaling to 0 does not really makes sense and probably won't work. Since placement needs to be discarded.


linasm (approved):
Makes sense.

jeromefroe (approved):

PR #327, opened 2022-07-12
[api] schema: extendedOptions as untyped object

schallert:
`extendedOptions` are used to store unstructured metadata. This PR
ensures the OpenAPI schema reflects that.


robskillington (approved):
LGTM

PR #329, opened 2022-08-08
Clean up unused PVCs

faustuzas:
After shrinking the cluster, allocated PVCs for the removed pods are not deleted at the moment. 

This PR will add a step to the controller loop which will check each cluster's STS if there are any unused PVCs assigned and will delete them.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -51,6 +51,10 @@ import (
 
 	jsonpatch "github.com/evanphx/json-patch"
 	m3placement "github.com/m3db/m3/src/cluster/placement"
+	pkgerrors "github.com/pkg/errors"
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+
 	myspec "github.com/m3db/m3db-operator/pkg/apis/m3dboperator/v1alpha1"
 	clientset "github.com/m3db/m3db-operator/pkg/client/clientset/versioned"
 	samplescheme "github.com/m3db/m3db-operator/pkg/client/clientset/versioned/scheme"
@@ -61,9 +65,6 @@ import (
 	"github.com/m3db/m3db-operator/pkg/k8sops/podidentity"
 	"github.com/m3db/m3db-operator/pkg/m3admin"
 	"github.com/m3db/m3db-operator/pkg/util/eventer"
-	pkgerrors "github.com/pkg/errors"
-	"github.com/uber-go/tally"
-	"go.uber.org/zap"
 )
 
 const (
@@ -247,7 +248,14 @@ func (c *M3DBController) Run(nWorkers int, stopCh <-chan struct{}) error {
 	defer cancel()
 
 	c.logger.Info("waiting for informer caches to sync")
-	if ok := cache.WaitForCacheSync(stopCh, c.clustersSynced, c.statefulSetsSynced, c.podsSynced, c.pvcSynced); !ok {
+	ok := cache.WaitForCacheSync(
+		stopCh,
+		c.clustersSynced,
+		c.statefulSetsSynced,
+		c.podsSynced,
+		c.pvcSynced,
+	)
+	if !ok {
 		return errors.New("caches failed to sync")
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/controller/controller.go
```diff
@@ -1267,6 +1279,65 @@ func (c *M3DBController) getParentCluster(pod *corev1.Pod) (*myspec.M3DBCluster,
...
+		logger.Info("Deleting persistent volume",
+			zap.String("persistent_volume", claim.Spec.VolumeName),
+			zap.Bool("is_old_enough", isOldClaim))
+
+		//err = c.kubeClient.CoreV1().PersistentVolumes().Delete(ctx, claim.Spec.VolumeName, metav1.DeleteOptions{})
+		//if err != nil {
+		//	return fmt.Errorf("deleting persistent volume: %w", err)
+		//}
+
+		logger.Info("Deleting persistent volume claim",
```
mfesenko, line 1325: I'm not 100% sure, but I would expect that PVCs need to be deleted before PVs.

faustuzas: My concern is that if we delete PVC and then crash, we don't have any references to track back the PV and we are not cleaning it anymore. 

I will check if after deleting PV, PVC does not create a new PV for itself

mfesenko: Have you tried running these operations in this order manually? I thought that it will not allow deleting a PV if there's a claim for it, but I don't know for sure. This blog post seems to address this issue too: https://jhooq.com/k8s-delete-pv-pvc/

faustuzas: I will test this out

faustuzas: I see that we use `reclaimPolicy: Delete` for our storage classes, so PV will get deleted automatically

pkg/controller/controller.go
```diff
@@ -1267,6 +1289,113 @@ func (c *M3DBController) getParentCluster(pod *corev1.Pod) (*myspec.M3DBCluster,
 	return cluster, nil
 }
 
+// cleanupUnusedStorageForSts removes all PVCs and PVs for the stateful which are not used.
+// This method should be called only when the cluster is stable
+// and there are no ongoing deployments.
```
Antanukas, line 1294: term `deployment` is probably foreign to operator domain. Operator simply tries to reach certain cluster state based on specification we set on M3DB crd. I think stable here means stateful sets match CRD definition and all pods are ready.

pkg/controller/controller.go
```diff
@@ -1267,6 +1289,113 @@ func (c *M3DBController) getParentCluster(pod *corev1.Pod) (*myspec.M3DBCluster,
...
+	}
+
+	pvcsToRemoveByName := make(map[string]*corev1.PersistentVolumeClaim, len(pvcs))
+	for _, pvc := range pvcs {
+		pvcsToRemoveByName[pvc.Name] = pvc
+	}
+
+	// first reclaim the PVCs which should be in use deciding by the parsed pod id
+	// and the replica count of the STS
+	lastPodIDInSTS := int(*sts.Spec.Replicas - 1)
```
Antanukas, line 1321: maybe it would be safier to iterate all pod specs in a statefulset, take PVC name from them and filter out from "all pvc list". After that you will be left with pvcs not belonging to any pods.

faustuzas: there is this check below

pkg/controller/controller.go
```diff
@@ -1267,6 +1289,113 @@ func (c *M3DBController) getParentCluster(pod *corev1.Pod) (*myspec.M3DBCluster,
...
+			PersistentVolumes().
+			Get(ctx, pvc.Spec.VolumeName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("getting k8s pv: %w", err)
+		}
+
+		// Sometimes a PV can have a reclaim policy other than "Delete". Skip these PVCs,
+		// so we would not leave dangling PVs which are hard to select because of the missing labels.
+		if pv.Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimDelete {
+			logger.Warn(
```
Antanukas, line 1369: We have seen glitches when PVs are not reclaimed even if they are "Delete" and pvc gets deleted. Maybe we should do "DeleteIfExists" for PV regardless if delete cascades or not.

faustuzas:
It was decided to move PVC cleaning to manual operations.

PR #331, opened 2022-08-19
[*] Include namespace info in more log messages

schallert:
Helps debugging to have the namespace / name of the cluster in all log
messages.


cerkauskas (approved):
LGTM

PR #333, opened 2022-08-20
Fix flaky tests

ryanhall07:
This fixes numerous flaky tests. Most of the flakes were caused because
the Indexer is updated async.


pkg/controller/controller.go
```diff
@@ -907,6 +907,12 @@ func (c *M3DBController) updateStatefulSetPods(
 			if err := c.kubeClient.CoreV1().
 				Pods(pod.Namespace).
 				Delete(ctx, pod.Name, metav1.DeleteOptions{}); err != nil {
+				if kerrors.IsNotFound(err) {
+					// This pod was already deleted in a previous iteration of the reconile loop. This can happen
```
ryanhall07, line 915: This was a real issue in the reconcile loop, due to the indexer being stale.

1. Delete Pod
2. Loop again and list pods. Pod still in index.
3. Indexer updated.
4. Delete Pod again but no longer in Indexer.

pkg/controller/controller_test.go
```diff
@@ -180,12 +181,14 @@ func waitForStatefulSets(
 	cluster *myspec.M3DBCluster,
 	verb string,
 	opts waitForStatefulSetsOptions,
-) (waitForStatefulSetsResult, bool) {
+	until func(waitForStatefulSetsResult) bool,
+) waitForStatefulSetsResult {
```
ryanhall07, line 185: since the indexer can take "awhile" to update, 10 iterations of the reconcile loop might not be enough. Instead of increasing this arbitrary number further, I changed the logic to provide condition to wait until.

pkg/controller/update_cluster.go
```diff
@@ -515,6 +515,8 @@ func (c *M3DBController) expandPlacementForSet(
 	if err != nil {
 		return err
 	}
+	// give a deterministic order to the pods for easy testing with gomock comparisons.
```
ryanhall07, line 518: fixes a gomock assertion flake

pkg/controller/update_cluster.go
```diff
@@ -532,12 +534,25 @@ func (c *M3DBController) expandPlacementForSet(
 		}
 	}
 	if len(podsToAdd) == 0 {
-		return errors.New("could not find pod absent from placement")
+		c.logger.Info("pods missing from placement are not indexed yet.")
```
ryanhall07, line 537: again another indexer delay that might be a real issue in prod.

pkg/k8sops/m3db/statefulset.go
```diff
@@ -178,7 +178,7 @@ func NewBaseStatefulSet(ssName, isolationGroup string, cluster *myspec.M3DBClust
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        ssName,
 			Labels:      objLabels,
-			Annotations: objAnnotations,
+			Annotations: annotations.Copy(objAnnotations),
```
ryanhall07, line 181: the sts and the pod were sharing the same annotations map, so when the sts annotations were updated with the operator annotation, the pod would inherit it as well.

robskillington: Interesting - do you think anyone/we may be relying on this behavior even if only accidentally for things like monitoring annotations?

Can probably fix it later with a cluster config update if that’s the case but definitely some kind of behavior i could see someone having started depending on.

robskillington (approved):
LGTM

pkg/k8sops/m3db/statefulset.go
ryanhall07: so I think this currently is a test only bug, since the test generates a stateful set _and_ sets the update annotation for the operator to update it. so the shared annotations are mutated and "corrupted".

in a real env, the annotation is set on an existing sts in k8s, so it doesn't hit this bug.

pkg/controller/update_cluster.go
schallert: One nice benefit of returning an error here is that it'll immediately trigger a requeue of the cluster event loop. If we return nil, it's likely that the event that causes the indexer to pick up the pod will trigger a reconcile, but I'm not 100% sure.

The sooner we re-reconcile the sooner the cluster converges on the desired state. Otherwise you have to wait for the requeue timeout (5min), or an unrelated event.

schallert (approved):
LGTM pending comment about error vs. not.

pkg/controller/update_cluster.go
ryanhall07: ah good point. I'll change it back to an error.

I guess the tests will just need to allow these of transient errors.

